PR #12, opened 2026-02-14
fix: prevent colgrep from loading wrong ONNX Runtime library

### LINKED ISSUE #9: colgrep crashes (exit code 101) during indexing on macOS 26 / Apple Silicon M3 Max
## Summary

`colgrep` crashes with exit code 101 (Rust panic) during the "Building index..." phase on macOS with Apple Silicon. This happens on every directory, including a single-file test directory. No error message is printed — only the model and "Building index..." lines appear before the crash.

## Environment

| Component | Version |
|-----------|---------|
| **ColGrep** | v1.0.7 (installed via curl installer) |
| **CPU** | Apple M3 Max |
| **OS** | macOS 26.0 (Tahoe), Build 25A354 |
| **Kernel** | Darwin 25.0.0 (arm64) |
| **Binary** | Mach-O 64-bit executable arm64 |

## Reproduction

```bash
# Minimal reproduction — single Python file
mkdir -p ~/colgrep-test
echo 'def log_error(msg): print(f"ERROR: {msg}")' > ~/colgrep-test/test.py
colgrep "log an error" ~/colgrep-test/
```

## Output

```
🤖 Model: lightonai/LateOn-Code-edge
📂 Building index...
```

Exit code: 101 (Rust panic, no backtrace even with `RUST_BACKTRACE=full`)

## Attempted Workarounds (all produce same crash)

| Attempt | Result |
|---------|--------|
| `colgrep settings --parallel 1` | ❌ Same crash |
| `colgrep settings --fp32` | ❌ Same crash |
| `colgrep settings --int8` | ❌ Same crash |
| `RUST_BACKTRACE=full` | ❌ No backtrace printed |
| `ORT_LOG_LEVEL=verbose` | ❌ No additional output |
| Different directories (small and large) | ❌ Same crash |
| `colgrep clear` + retry | ❌ Same crash |

## Analysis

- The model downloads successfully and is cached at `~/.cache/huggingface/hub/models--lightonai--LateOn-Code-edge/`
- Model files present: `model.onnx`, `model_int8.onnx`, `tokenizer.json`, configs
- The crash happens at the ONNX inference stage (after parsing, during embedding)
- Links against `Accelerate.framework` — may be a CoreML/ANE issue on the M3 Max or macOS 26 (Tahoe)
- No backtrace suggests the panic may be in the ONNX C runtime rather than Rust code

## Expected Behavior

`colgrep` should successfully build the index and return search results.

joe32140:
## Summary

Fixes #9 — colgrep crashes with silent exit code 101 during indexing.

**Root cause**: `find_onnx_runtime()` in `onnx_runtime.rs` uses a filename pattern that matches the wrong library. On systems with conda/pip-installed onnxruntime, the directory contains both:

| File | `starts_with("libonnxruntime")` | `ends_with(".so")` | Matched? | Has `OrtGetApiBase`? |
|---|---|---|---|---|
| `libonnxruntime.so.1.23.2` | ✓ | ✗ (ends `.2`) | **skipped** | ✓ |
| `libonnxruntime_providers_shared.so` | ✓ | ✓ | **selected** | ✗ |

The correct library is right there but the pattern misses it (Linux versioned `.so` files end with the version number). The only file matching the pattern is a companion library that lacks `OrtGetApiBase`, causing `ort` to panic.

**Fix 1** — Tighten the filename pattern:
```
- starts_with("libonnxruntime") && ends_with(".so")
+ starts_with("libonnxruntime.so")
```
This matches `libonnxruntime.so` and `libonnxruntime.so.1.23.2` but not `libonnxruntime_providers_shared.so`.

**Fix 2** — Don't swallow panics during model loading:

The `with_suppressed_stderr()` wrapper (for CoreML warnings on macOS) redirects stderr to `/dev/null`. When `ort` panics, the panic message is lost — the user sees only `exit code 101`. Wrapping in `catch_unwind` lets the panic re-raise after stderr is restored.

## Test plan

- [x] Reproduced on Linux x86_64 (WSL2) with `CONDA_PREFIX` set and onnxruntime installed via pip in base conda env
- [x] Verified fix resolves the crash — colgrep indexes and searches successfully
- [x] Verified the correct library (`libonnxruntime.so.1.23.2`) is now selected
- [x] Verified panic messages are visible when stderr suppression is active

### DIFF
--- a/colgrep/src/index/mod.rs
+++ b/colgrep/src/index/mod.rs
@@ -232,16 +232,26 @@ impl IndexBuilder {
                 .unwrap_or_else(crate::config::get_default_batch_size);
 
             // Suppress stderr during model loading to hide CoreML's harmless
-            // "Context leak detected" warnings on macOS
-            let model = crate::stderr::with_suppressed_stderr(|| {
-                Colbert::builder(&self.model_path)
-                    .with_quantized(self.quantized)
-                    .with_parallel(num_sessions)
-                    .with_batch_size(batch)
-                    .with_execution_provider(execution_provider)
-                    .build()
-            })
-            .context("Failed to load ColBERT model")?;
+            // "Context leak detected" warnings on macOS.
+            // Use catch_unwind so that if ONNX Runtime panics (e.g., wrong dylib),
+            // the error message is still shown to the user after stderr is restored.
+            let model_result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+                crate::stderr::with_suppressed_stderr(|| {
+                    Colbert::builder(&self.model_path)
+                        .with_quantized(self.quantized)
+                        .with_parallel(num_sessions)
+                        .with_batch_size(batch)
+                        .with_execution_provider(execution_provider)
+                        .build()
+                })
+            }));
+            let model = match model_result {
+                Ok(result) => result.context("Failed to load ColBERT model")?,
+                Err(panic_payload) => {
+                    // Re-panic with the original payload now that stderr is restored
+                    std::panic::resume_unwind(panic_payload);
+                }
+            };
 
             self.model = Some(model);
         }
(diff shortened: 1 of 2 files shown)


raphaelsty (approved):
LGTM, thank you, congratulations for the PR

PR #20, opened 2026-02-19
fix(colgrep): apply paths ignore based on relative paths, not absolute ones

### LINKED ISSUE #19: `colgrep init` indexes 0 files when the project root is inside a hidden directory
When running `colgrep init` from inside a dot-prefixed directory (e.g. `~/.pi/agent/extensions/`), it indexes 0 files because the hidden-directory filter applies to the resolved project root path itself, not just to discovered subdirectories.

**Repro:**

```bash
mkdir -p ~/test/.dotdir
echo 'export function hello() { return "world" }' > ~/test/.dotdir/index.ts

colgrep init ~/test/.dotdir
# => Index is up to date for /Users/me/test/.dotdir (0 files)

colgrep init ~/test
# => Indexed /Users/me/test (added: 0) — .dotdir skipped, as expected
```

**Expected:** When the user explicitly passes a hidden directory as the project root (or runs `colgrep init .` from inside one), the root should be trusted and files within it indexed. The hidden-directory filter should only apply to subdirectories discovered during the walk, not to ancestors of/the project root itself.

**Use case:** Tools like [pi](https://github.com/mariozechner/pi-coding-agent) store config and extensions in `~/.pi/agent/extensions/`. Running colgrep there to search extension source code doesn't work.

marchellodev:
Fixes #19

### DIFF
--- a/colgrep/src/index/mod.rs
+++ b/colgrep/src/index/mod.rs
@@ -643,7 +643,9 @@ impl IndexBuilder {
             }
 
             // Skip files in ignored directories (same filtering as scan_files)
-            if should_ignore(&full_path) {
+            // Use the relative path so hidden-directory filtering doesn't reject
+            // ancestor components of the project root itself.
+            if should_ignore(path) {
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


raphaelsty:
LGTM, congratulations and thank you for this fix 😊

Issue #34, opened 2026-03-12
Windows CUDA bring-up for colgrep: strict GPU execution, env/linker fixes, and VRAM usage


cepera-ang:
So, I'm trying to run `colgrep` on Windows with GPU (CUDA) enabled and so far it was quite a journey.

## Goal

Make this work reliably on Windows with CUDA 13.1:

```powershell
cargo run --bin colgrep --features cuda -- init
```
and make it actually use CUDA rather than silently falling back to CPU to index large codebases.

I hit several different classes of failure in sequence.

### 1. `cudarc` API and dependency mismatch

`next-plaid` depended on `cudarc 0.12`, which is very old and doesn't support the current CUDA (13.1 in my case). So I updated to to `cudarc 0.19` both in `next-plaid` and `fastkmeans-rs`, which required some straightforward changes in code (will make a PR there if you want them)
`cudarc 0.19` changed the safe driver API shape:

- `CudaDevice` -> `CudaContext` plus `CudaStream`
- stream-based copies and launches
- module/function loading API changes

fix: 
- upgraded `next-plaid` to `cudarc 0.19.3`
- patched `fastkmeans-rs` to `cudarc 0.19.3`
- updated both CUDA wrappers to the `0.19` API

Without that migration, the CUDA path did not build.

### 2. `ort` builder errors in `next-plaid-onnx`

The ONNX builder code used `context(...)` and `?` on builder methods returning `ort::Error<SessionBuilder>`.

Problem:

- those errors no longer satisfied the trait bounds expected by `anyhow::Context`

Fix:

- replaced the failing `context(...)`/`?` sites with inline `map_err(...)`

Why:

This was the smallest source change that preserved behavior and made the ONNX builder compile again.

### 3. Windows linker could not find `cuda.lib`

At that point the build failed with:

```text
LINK : fatal error LNK1181: cannot open input file 'cuda.lib'
```
Fix:

- set user-level environment variables:

```text
CUDA_PATH=C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v13.1
LIB=C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v13.1\lib\x64
```

### 5. `colgrep` compiled and ran, but did not actually use CUDA

The binary started, but `nvidia-smi` showed no real GPU work.

Root cause:

`colgrep` still had CPU-preferring logic before ONNX was even asked to create a CUDA session.

Specifically:

- small batches forced CPU
[...]

PR #36, opened 2026-03-14
Bring up CUDA-enabled colgrep on Windows

cepera-ang:
As discussed in #34 this PR is an attempt to make sure that CUDA version of colgrep runs end to end on GPU without any unexpected fallbacks.

1. It updates cudarc to 0.19.3 and fixes calls to new API.
2. Forces GPU usage if build with CUDA feature (ignoring small batches, etc)
3. Assumes cudnn is available and configured correctly (env variables set)
4. releases and reloads a model at each 1000 iterations to conserve VRAM (otherwise it goes OOM on 8GB with 17M model) 

Requires https://github.com/lightonai/fastkmeans-rs/pull/2 to land first (and will need an update to a new version after that, I used vendored version locally for testing).


raphaelsty:
Amazing @cepera-ang then I will create another pull request based on yours, I will add you as co-author and merge your update, I might update few things or two but I'll keep what make colgrep working fine on windows with careful testing for other OS

raphaelsty:
I think we can close this MR now windows is supported for cuda :)

PR #64, opened 2026-03-26
Optimize query latency on large repos (~4x faster)

raphaelsty:
## Summary

- Reduces query time on PyTorch (240K code units, 3.1GB index) from ~5s to ~1.2s (warm, CPU) by eliminating redundant per-query overhead
- Uses mtime-based fast path to skip content hashing when files haven't changed
- Gates expensive cleanup_orphaned_entries behind deletion detection or periodic trigger
- Adds dirty flag to IndexState to skip repair_index_db_sync on clean shutdown
- Caches query embeddings and FTS5 results across the dual-search path (encode once, search twice)
- Adds fast path in MmapIndex chunk merging to skip scanning all 120 chunk files when manifest is valid

## Test plan

- [x] `cargo test` passes (49 colgrep tests, 91 next-plaid tests, 29 next-plaid-onnx tests)
- [x] `cargo test --test filtering_integration` passes (9 tests including update/delete workflow)
- [x] Verified result correctness: `-e "def forward" "neural network" -k 10 -l` matches baseline exactly
- [x] Verified performance on PyTorch repo with `time colgrep "error handling" -k 5 -l`
- [ ] Test on repos with active file churn to verify mtime fast path correctly detects changes
- [ ] Test interrupt recovery: dirty flag should trigger repair on next run

### DIFF
--- a/colgrep/src/index/mod.rs
+++ b/colgrep/src/index/mod.rs
@@ -1166,7 +1166,8 @@ impl IndexBuilder {
         // 0. Clean up orphaned entries (files in index but not on disk)
         // This is expensive on large repos (fetches all metadata from SQLite),
         // so only run when deletions were detected or periodically as a safety net.
-        let should_cleanup = !plan.deleted.is_empty() || old_state.search_count.is_multiple_of(50);
+        let should_cleanup = !plan.deleted.is_empty()
+            || (old_state.search_count > 0 && old_state.search_count.is_multiple_of(50));
         let orphaned_deleted = if should_cleanup {
             self.cleanup_orphaned_entries(index_path_str)?
         } else {
@@ -3035,9 +3036,14 @@ impl Searcher {
 /// Fix SQLite type conversions in metadata JSON.
 ///
 /// SQLite stores booleans as integers (0/1) and arrays/objects as JSON strings.
-/// This function detects and converts them back generically:
-/// - Integer 0/1 with a `has_` or `is_` prefix → boolean
+/// This function detects and converts them back using naming conventions:
+/// - Integer fields with a `has_` or `is_` prefix → boolean
 /// - String values that parse as JSON arrays → array
+///
+/// Note: this is more general than the original hardcoded field list — it will
+/// convert *any* `has_`/`is_` integer field to bool, not just known ones.
+/// This is intentional: new metadata fields following the naming convention are
+/// handled automatically without code changes.
 fn fix_sqlite_types(meta: &mut serde_json::Value) {
     if let serde_json::Value::Object(ref mut obj) = meta {
         let keys: Vec<String> = obj.keys().cloned().collect();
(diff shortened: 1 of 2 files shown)


raphaelsty (commented):
## Review: Optimize query latency on large repos (~4x faster)

**Overall**: Well-structured optimization PR. The changes are well-motivated — eliminating redundant per-query overhead (encoding, FTS5, file scanning, chunk validation) is the right approach. Two commits, both with clear messages. +262/-70 across 4 files.

### Correctness

**Dirty flag lifecycle** — Correct. Set before writes, cleared after success, preserved on interrupt (bail without save). Repair only runs when the flag is set. Clean design.

**Mtime fast path in `compute_update_plan`** — Safe. Uses exact `f64` equality on file mtime, but falls through to content hashing on mismatch, so it's a pure performance optimization with no correctness risk. A file modified twice within the same fractional second would still be caught by the hash.

**FTS5 result caching** — Correct. `search_hybrid_with_embedding` (lines 2960-2976) filters pre-computed FTS5 results to the current subset via a `HashSet` intersection when the subset differs. Well handled.

**Merge chunk fast path** — Robust. Validates: num_chunks, padding_rows, manifest consistency, metadata.json mtime (with epsilon), merged file existence, and exact file size. The epsilon fix in the second commit addresses serde f64 rounding — good catch.

### Suggestions

1. **`fix_sqlite_types` extraction changed behavior** (minor). The original code hardcoded specific field names (`has_loops`, `has_branches`, `calls`, etc.). The new version uses heuristics: `has_`/`is_` prefix for booleans, `starts_with('[')` for arrays. This is more general but subtly different — it will now convert *any* `has_`/`is_` integer field to bool, and any JSON-encoded string starting with `[` to an array. This is likely fine (and arguably better), but it's a behavioral change beyond pure optimization. Worth noting in the PR description.

2. **`is_multiple_of(50)` on `search_count`** — When `search_count == 0` this returns `true` (0 is a multiple of everything), so orphan cleanup runs on the very first search even with no deletions. Probably fine (first search is cold anyway), but if the intent was "every 50th search", consider `search_count > 0 && search_count % 50 == 0`.

3. **`npy_header_size_1d` / `npy_header_size_2d`** — These duplicate the header format logic from `write_npy_header_1d`/`write_npy_header_2d`. If the header format ever changes, these would need to stay in sync. A comment cross-referencing them would help, or you could refactor to share the size computation.

4. **Two unchecked test plan items** — The PR description has two unchecked items:
   - "Test on repos with active file churn to verify mtime fast path correctly detects changes"
   - "Test interrupt recovery: dirty flag should trigger repair on next run"
   
   These are worth completing before merge, especially the interrupt recovery test.

### Verdict

**Approve with minor comments.** The core optimizations are sound and well-guarded by fallbacks. The `fix_sqlite_types` behavior change and `is_multiple_of(0)` edge case are the only things worth addressing. Nice work on the 4x speedup.

Issue #10, opened 2026-02-13
Unable to index a large project (Swift compiler)


artemdinaburg:
# Summary

When indexing the Swift compiler source (which will also pull in LLVM source code, so it is quite a lot of code) I consistently hit the following error:

```
thread 'main' (121905455) has overflowed its stack
fatal runtime error: stack overflow, aborting
zsh: abort      colgrep init $(pwd)
```

# Environment

MacOS 26.2
colgrep 1.0.7 (installed today via `curl --proto '=https' --tlsv1.2 -LsSf https://github.com/lightonai/next-plaid/releases/latest/download/colgrep-installer.sh | sh`)

# Reproduction

Follow instructions to clone & build the Swift project at https://github.com/swiftlang/swift/blob/main/docs/HowToGuides/GettingStarted.md

Attempt to index it:
```
cd ~/git/swift-project
colgrep init $(pwd)
```

Output:

```
colgrep init $(pwd)
model_int8.onnx [00:00:01] [██████████████████████████████████████████████████████████████████████████████████████] 16.43 MiB/16.43 MiB 36.13 MiB/s (0s)
tokenizer.json [00:00:00] [█████████████████████████████████████████████████████████████████████████████████████████] 3.42 MiB/3.42 MiB 15.56 MiB/s (0s)
..fig_sentence_transformers.json [00:00:00] [███████████████████████████████████████████████████████████████████████████████████████] 762 B/762 B - (0s)
config.json [00:00:00] [██████████████████████████████████████████████████████████████████████████████████████████████████████] 1.22 KiB/1.22 KiB - (0s)
onnx_config.json [00:00:00] [███████████████████████████████████████████████████████████████████████████████████████████████████████] 792 B/792 B - (0s)
model.onnx [00:00:02] [███████████████████████████████████████████████████████████████████████████████████████████] 64.82 MiB/64.82 MiB 43.28 MiB/s (0s)
⠉ [▓░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░] 1551/113717 (3m) Parsing files...                                                                          ⚠️  Skipping /Users/artem/git/swift-project/llbuild/thirdparty/protobuf/php/src/GPBMetadata/Google/Protobuf/Timestamp.php (stream did not contain valid UTF-8)
[...]

PR #67, opened 2026-03-30
Performance Exploration & Maximum Pipelining

cepera-ang:
So, I was tinkering with colgrep and noticed that there is a lot of performance left on the table. This was specifically focused on GPU path with next-plaid stuff running on CPU because the version I ran haven't been updated to support installed CUDA 13.1. All tests were done on small colgrep codebase (~2300 code units) and Zed (~60400 code units). I was doing and un-doing a lot of stuff, random and purposeful but it's all a bit faded away in my memory, so that write up is a reconstructed narrative of things I ended up keeping and 

Initial results: Zed codebase index in 1520seconds, next-plaid codebase in 25s.

## The Overarching Idea

It's simple: we have several stage, each of them takes different resources: embeddings are encoded on GPU, pooled, clustered and coded on CPU and written to disk, utilizing I/O. And in theory it should be possible to finish in time it takes for the longest stage to complete + finishing last batches and some startup overhead. 

Quick debug prints showed that the longest stage is GPU encoding and everything else may more or less finish while GPU is busy. Theory of constraints tells us that we should do everything possible to keep the bottleneck busy at all times. The result is a system where parsing, tokenization, GPU inference, pooling, vector-index construction, and metadata-DB writes all overlap in time — what used to be a sequential `encode → index → write-metadata` loop is replaced by a concurrent pipeline with bounded channels between stages.

## 1. Parallelised Call-Graph Construction (`colgrep/src/parser/call_graph.rs`)

The startup on Zed codebase was very slow between the first progress bar and the second, it took a minute with almost no load anywhere, it just sit there. Turns out, it was single threaded and quick `par_iter()` made that instantly 20x faster. No longer noticeable but at the end, I thought that in principle this could become another stage in the pipeline for really large codebases.  The inner loop over `(caller, calls)` pairs is now `par_iter()` over the calls map, collecting `(callee_idx, caller_name)` edges into a flat vector, then merging them into `called_by` maps sequentially. For large codebases with thousands of cross-references this parallelises the O(calls × avg_callees) work.
- [ ] Think about making that also a stage of pipeline to start further work earlier on really huge codebases.

---
## 2. Metadata DB — Fixed Schema & Transactional Batching (`next-plaid/src/filtering.rs`)

Another small win: I spotted that metadata writing at the end of each chunk took suspiciously long time (about 5-8s, 5 minutes total for Zed codebase) for what it should do (final metadata.db size is ~550MB so the lower bound to write it at once should be about a second on modern nvme). I checked what it's doing and it was using sqlite as a kind of nosql database updating the table schema at every row insert and every row was inserted in a separate transaction. So, I added fixed schema happy path and batched inserts of all rows in a single transaction.  

1. **Fixed-schema happy path**: Instead of per-batch column inference + `ALTER TABLE` churn, a schema inferred from the first row and used optimistically. `colgrep` always does the same rows so it eliminates the `PRAGMA table_info` + column-diff overhead entirely.

2. **Transaction wrapping**: All `INSERT` loops (in both the old generic path and the new fixed path) now run inside an explicit `BEGIN/COMMIT` transaction. SQLite's default auto-commit mode issues a separate `fsync` per `INSERT`; wrapping in a transaction reduces this to one `fsync` per chunk.

3. **`configure_connection`** applies `PRAGMA journal_mode=WAL; synchronous=NORMAL; temp_store=MEMORY;` uniformly across all paths.

## 3. Code unit Preparation and Tokenizer (`colgrep/src/embed.rs`)

Next, I noticed that metadata written to DB is enormous (largest code unit from Zed is 440kB) and there is no way model can process that. And yeah, maximum length is 2048 tokens, but we still tokenized all these megabytes of code and immediately truncated the tokenized documents to the maximum allowed. This is the place I really need input from @raphaelsty on what was the original intent. For now I just truncated the code to safe limit 8K chars (actually 6k is always enough to produce 2k tokens)
- **Truncation at 8K characters**: `MAX_EMBEDDING_TEXT_CHARS = 8192` with a `[...truncated...]` marker. 
- **File path moved before code** in the embedding text, so even when truncation kicks in, the file-provenance context is always preserved.
- [ ] Clarify what was the original intent for these huge code blocks and how should we really deal with them (split further?)
Further investigation showed that tokenizer automatically uses rayon pool and unnecessarily parallelizes work when overhead from parallelization dominates and we can basically finish all the tokenization in one thread and keep gpu encode fed just fine, so I made a dedicated rayon pool for tokenizer (or you need to mess with env variables to ask it to be single threaded)

## 4. The Pipelined Architecture (`colgrep/src/index/mod.rs`)

**~1400-line rewrite** of the indexing path, replacing the old sequential batch loop with a multi-stage pipeline connected by `mpsc` channels:

```
Parse → Dedup → [chan] → Tokenize → [chan] → GPU Encode → [chan] → Pool → [chan] → Index+Metadata
```

### Key pipeline stages and types

| Stage                | Thread               | Input type                   | Output type               | Bounded?             |
| -------------------- | -------------------- | ---------------------------- | ------------------------- | -------------------- |
| **Dedup**            | main                 | `(&CodeUnit, String)` chunks | `PreparedChunk`           | unbounded            |
| **Tokenize**         | `run_tokenize_stage` | `PreparedChunk`              | `TokenizedChunk`          | unbounded            |
| **GPU Encode**       | `run_encode_stage`   | `TokenizedChunk`             | `RawEncodedChunk`         | unbounded            |
| **Pool**             | `run_pool_stage`     | `RawEncodedChunk`            | `PooledChunkForIndex`     | `SyncSender` (cap 4) |
| **Index + Metadata** | `run_index_stage`    | `PooledChunkForIndex`        | `IndexedChunkForMetadata` | `SyncSender` (cap 8) |

The capacities (`POOLED_EMBEDDING_QUEUE_CAPACITY = 4`, `METADATA_QUEUE_CAPACITY = 8`) are guessed so:
- The tokenizer can race ahead all the way, filling a buffer for the GPU.
- The GPU stage is never waiting for tokenized batches nor stops working when downstream struggles.
- Pooling and indexing run in parallel, with enough buffer to smooth out per-chunk time variance.

### Embedding text deduplication

This one is surprised me, after sorting texts by size I noticed huge amount of "}" fragments. They are generated by some upstream processing that ensures that the code files are fully covered. Obvious idea is to embed them once and set the same embedding to every such chunk. Turns out, there are about 4700 duplicated chunks out of 60400, so we can instantly save 7-8% of GPU work. So, before hitting the GPU, each chunk now deduplicates embedding texts (`prepare_deduplicated_chunk`). Duplicate code units share a single encoding, and a `original_to_unique` index array maps results back. For codebases with many duplicated functions/method signatures, this cuts GPU work proportionally.

### K-means / codec pipelining on initial create

When no index exists yet, the first chunk's embeddings kick off a **separate k-means thread** (`colgrep-kmeans`) that computes centroids and codec artifacts concurrently while the next chunks continue encoding. The pipeline stage (`run_index_stage`) waits for the k-means handle only when it's ready to write the first chunk to disk — by which time k-means has usually finished.

### Sorted-greedy-pack batching & sort order control

New `EncodeSortOrder` enum (`BigFirst` / `SmallFirst`) and CLI flags `--batch-sort-order`, `--encode-batch-size`, `--index-chunk-size` give explicit control over how documents are grouped and ordered before hitting the GPU. The default `SmallFirst` puts small documents together quickly chew through the most of the code units early (like more than half of the progress in the first minute). Sorting is important because without it there is an  enormous amount of padding that is empty but still gets computed on GPU as if it is real data (i.e. one len=2048 document and 15 len=1 will run as slow as 16 x 2048, hundred times slower). 

---
## 5. ONNX Embedding Batch Planning (`next-plaid-onnx/src/lib.rs`)

The problem here is that we were still underutilizing GPU: the models are tiny and batch-size selected for the maximum document length meant that early small code units (majority of the codebase) weren't enough to bring high utilization. I spent a lot of time here, even making a VRAM-aware batch planner querying system for available free VRAM and inspecting the model and optimally packing documents into batches with sorted-greedy-pack algorithms. In the end, the simple heuristic `token_budget = configured batch_size x maximum document length` and fixed set of allowed batch sized worked fine. 

The **`tokenize_documents_in_batches()`** method uses this budget to split documents into micro-batches. Padding lengths are rounded up to 32-token quanta (`round_up_len_for_planning`) so ONNX runtime can reuse cached execution plans.

# Below is an AI summary of further changes (these I asked to make at some point, but didn't try to fully understand or test)

---
### 5b. Streaming document encoder

New `encode_documents_stream()` and `encode_documents_raw_stream()` methods return `Iterator`-backed streaming types (`DocumentEmbeddingStream` / `RawDocumentEmbeddingStream`) that drive **one worker thread per ONNX session**, pulling from a shared `VecDeque` work queue under a mutex. This means multi-session GPU batches can overlap in time, and the consumer (the pipeline index stage above) can begin writing chunks before the entire corpus is encoded.

### 5c. Separated tokenize/encode path

`prepare_batch_for_session()` and `encode_prepared_batch_with_session()` now exist as distinct public functions, letting the colgrep pipeline run tokenization on a dedicated thread while the ONNX session thread remains free for pure GPU work.

### Other ONNX changes

- `Colbert` struct fields wrapped in `Arc` (sessions, config, skiplist) to support `Clone` and cross-thread sharing.
- `pool_embeddings_hierarchical` now uses pre-allocated `Array2` + `cluster_sums`/`cluster_counts` arrays instead of per-cluster `Vec<Vec<f32>>` → eliminates scattered heap allocations in hot pooling loop.
- `pool_document_embeddings()` extracted as a public function so the pipeline pool stage can call it independently.

---

## 6. Codec — Parallel Batch Compression (`next-plaid/src/codec.rs`)

`compress_into_codes` replaced its serial `for start in 0..n step batch_size` accumulator loop with:

```
batch_ranges → into_par_iter() → map (GEMM + argmax per batch) → collect → flatten
```

- Batches are now processed by **rayon parallel iterator**, so on a 16-core machine all batches of the `[batch × centroids]` score matrix multiplications happen concurrently.
- Per-row argmax is now a plain sequential `.map()` (no nested `into_par_iter`) — eliminates nested-parallelism overhead that was causing rayon work-stealing contention.
- Default memory cap lowered from **4 GB → 1 GB** (`DEFAULT_MAX_NEAREST_CENTROID_MEMORY`), configurable at runtime via `NEXT_PLAID_MAX_NEAREST_CENTROID_MEMORY_MB`. The lower cap trades slightly smaller batches for dramatically reduced page-fault & zero-fill overhead from giant temporary score buffers.
- Batch size clamped to 1024 (down from 2048) for the same reason.

---

## 7. Outlier Detection — Blocked Tiling Kernel (`next-plaid/src/update.rs`)

The old `find_outliers` GEMM-based approach allocated an `[n × k]` similarities matrix, which was O(GB) for large centroid counts. The new `find_outliers_blocked` replaces it with a **tiled, allocation-free kernel**:

- Centroids are processed in tiles of `OUTLIER_CENTROID_TILE = 8`.
- Embeddings are processed in blocks of `OUTLIER_EMBEDDING_TILE = 64`.
- Within each tile×block, a fixed-size `dots[64][8]` accumulator is filled directly from flat raw slices — no ndarray iterators, no intermediate heap allocations.
- Pre-computed `centroid_norms_sq` and `emb_norms` (using a 4-wide manual unrolled `squared_norm`) enable `||a-b||² = ||a||² + ||b||² − 2·a·b` distance with a single dot-product pass.
- **Borderline re-check**: embeddings within `OUTLIER_THRESHOLD_RECHECK_REL_EPS = 1e-5` of the threshold are re-computed using `f64` precision (`min_distance_sq_precise`) to avoid false positives from `f32` catastrophic cancellation.
- Parallel over embedding blocks via `par_chunks` with `with_min_len(4)`.
- The old GEMM version is kept as `find_outliers_gemm` (gated behind `#[cfg(test)]`) so tests can cross-validate results.
- `select_outlier_gemm_batch_size` snaps batch sizes to a fixed power-of-two ladder for stable profiling.

This is much faster and I checked that the results are identical on 99% of batches down to a single value sometimes being different due to floating point order differences. 

---

## 8. Index Construction — Split Into Composable Stages (`next-plaid/src/index.rs`)

The monolithic `create_index()` function is now decomposable into three public building blocks:

1. **`prepare_codec_artifacts(embeddings, centroids, config)`** — computes heldout residuals, bucket cutoffs/weights, avg residual per dim, and cluster threshold. Returns `PreparedCodecArtifacts`.
2. **`encode_index_chunk(embeddings, codec, force_cpu)`** — encodes one chunk of doc embeddings into codes + packed residuals. Returns `EncodedIndexChunk`. Handles CUDA-with-fallback internally.
3. **`write_index_from_encoded_chunks(chunks, codec_artifacts, index_path, config)`** — writes all `.npy`, `.json`, IVF, and metadata files to disk. Returns `Metadata`.

This decomposition is what enables the pipeline above: the colgrep pipeline can call `encode_index_chunk` per-chunk as embeddings arrive, accumulate `EncodedIndexChunk`s, and call `write_index_from_encoded_chunks` once at the end.

---

## 9. Build & Runtime Configuration

- **Global rayon pool sizing** (`init_global_rayon_pool` in `colgrep/src/main.rs`): defaults to `available_parallelism - 4` (reserving cores for the pipeline threads), tunable via `NEXT_PLAID_GLOBAL_RAYON_THREADS`.
- New CLI flags: `--batch-size`, `--encode-batch-size`, `--index-chunk-size`, `--batch-sort-order` for explicit benchmarking control.
- [ ] There is a need for some tracing infrastructure, I used random ad hoc printing and `samply` for profiling but it is messy and unsustainable, good centralized tracing would be so much better.
- [ ] Another huge todo is to have some standardized set of good end to end tests: some big enough code base and the results we expect from queries on it (like some existing benchmark/eval?) 

# Finally

The result in the image. Orange bar 'colgrep-encode' more or less corresponds to GPU load and the rest is CPU load. We can see that everything nicely overlaps, GPU is fully fed and total time is 'gpu time + last chunk processing'. We can clearly see that initial chunks are processed extremely fast and the last 6 chunks full of 2048 len (truncated btw) documents take the same time as 55 chunks before them. 

Final time for Zed: 220s (200s GPU), final time for next-plaid: 12s (7s GPU)

I understand that dumping that huge change set is not the best idea, feel free to take ideas, code, ask for follow up or changes.

image


raphaelsty (commented):
Just leaving some comments, lot of work there, but this is amazing :)

raphaelsty:
Hi @cepera-ang, currently working on your MR, don't bother to update anything related to the comment, those are generated using Claude as an anchor, I'm working on it's integration step by step from other PR, you will be added as co-author of the MR each time, thank's a lot for your amazing work, this is really cool

raphaelsty:
## GPU Benchmark: `main` vs this PR (AI-generated)

Ran a single-pass A/B benchmark using `colgrep` with `FORCE_GPU=1` on an H100 80GB (CUDA), indexing and searching 15 Rust files (~12.5K lines) from `next-plaid/src/`.

### Results

| Metric | `main` | This PR | Delta |
|---|---|---|---|
| **Index build** | 19.5s | 4.8s | **-75% (4x faster)** |
| **Search (8 queries)** | 3.8s total | 14.9s total | **+297% (~4x slower)** |
| **Search avg/query** | 0.47s | 1.86s | **+297% (~4x slower)** |

### Analysis

**Index build is ~4x faster** — The ONNX pipelining changes in `next-plaid-onnx` (tokenize → encode → pool stages running in parallel threads, dynamic batching by sequence length, `with_memory_pattern` kept enabled for GPU) are clearly paying off for bulk document encoding.

**Search is ~4x slower** — The root cause is the change to the `Searcher` initialization in `colgrep/src/index/mod.rs`. On `main`, the searcher **always uses CPU** for query encoding:

```rust
// main branch (colgrep/src/index/mod.rs ~L2432)
// Load model for search - use CPU or CoreML (not CUDA)
// CUDA has too much overhead for single query encoding, CPU/CoreML is faster
let execution_provider = if cfg!(feature = "coreml") {
    ExecutionProvider::CoreML
} else {
    ExecutionProvider::Cpu
};
```

On this PR, when `FORCE_GPU=1` is set, the searcher now uses CUDA for query encoding:

```rust
// PR branch
let execution_provider = match acceleration_mode {
    AccelerationMode::ForceGpu => ExecutionProvider::Cuda,
    AccelerationMode::ForceCpu => ExecutionProvider::Cpu,
    AccelerationMode::Auto => { /* CoreML > CPU */ }
};
```

For single-query encoding (1 short text), CUDA has significant overhead: kernel launch latency, CPU↔GPU memory transfers, and ONNX Runtime CUDA EP initialization per session — all of which dwarf the actual compute for a single query. The original comment on `main` was correct: "CUDA has too much overhead for single query encoding."

Additionally, the PR now runs FTS5 hybrid search by default (`search_hybrid` → `fts5_search` + RRF fusion), adding an SQLite FTS5 query on every search call. This adds latency compared to pure semantic search on `main`.

### Suggestion

For the `Auto` acceleration mode, consider keeping the `main` branch behavior of using CPU for query-time encoding while using GPU only for bulk document encoding (indexing). The `ForceGpu` mode could still honor the GPU path for users who explicitly want it, but with a note that single-query latency will be higher. The FTS5 hybrid search overhead could also be investigated to see if the SQLite query can be made cheaper or run in parallel with the vector search.

---
*This comment was AI-generated using Claude Code. Benchmark was run on a single NVIDIA H100 80GB with `CUDA_VISIBLE_DEVICES=2`.*

raphaelsty (commented):
## PR Review: Performance Exploration & Maximum Pipelining

Overall, this is a well-structured effort to pipeline the indexing stages (parse → tokenize → encode → pool → index → metadata) so they overlap in time. The concurrent architecture is sound. Below are specific findings, organized by severity and category.

---

### Memory Concerns

**1. Unbounded channels in the pipeline can cause memory blowup**
`colgrep/src/index/mod.rs` — The first three pipeline stages use unbounded `mpsc::channel`:
```rust
let (tokenize_tx, tokenize_rx) = mpsc::channel::<PreparedChunk>();
let (encode_tx, encode_rx) = mpsc::channel::<TokenizedChunk>();
let (pool_tx, pool_rx) = mpsc::channel::<RawEncodedChunk>();
```
While the downstream stages (index, metadata) use bounded `sync_channel` with capacities of 4 and 8, the upstream stages are unbounded. If tokenization is fast but GPU encoding is slow (the expected bottleneck), `TokenizedChunk` messages — each containing full token tensors — will accumulate without bound. For a 30K+ unit codebase, this could mean hundreds of MB of intermediate data queued in memory.

**Recommendation:** Replace with `mpsc::sync_channel()` using small bounded capacities (2–4), consistent with the downstream stages. This provides backpressure without reducing throughput.

---

**3. Parallel `compress_into_codes` multiplies peak scores-matrix memory**
`next-plaid/src/codec.rs` ~lines 304–332 — The old code processed batches sequentially, so only one `[batch_size, num_centroids]` scores matrix existed at a time. The new code uses `into_par_iter()`, so rayon may run 8–16 batches simultaneously, each allocating its own scores matrix. The `max_nearest_centroid_memory()` budget is per-batch, not total — the constant name and comment are now misleading. Consider either dividing the budget by the rayon thread count, or documenting this explicitly.



---

**4. `initial_create` path accumulates all `EncodedIndexChunk`s in memory before writing**
`colgrep/src/index/mod.rs` ~lines 340–355 — In the initial index creation path, the coding thread collects **all** encoded chunks into a `Vec` before writing. Each `EncodedIndexChunk` contains `codes: Array1<i64>` and `residuals: Array2<u8>`. For very large codebases (100K+ units), all compressed index data is held in memory at once. The old incremental `MmapIndex::update_or_create` path wrote each chunk immediately. Consider streaming writes or documenting the higher peak memory requirement.

---

**6. `run_pool_stage` clones full embedding matrices for deduplicated documents**
`colgrep/src/index/mod.rs` ~lines 265–271 — When deduplication occurs, the same `Array2<f32>` is `.clone()`'d for each duplicate. Each embedding is ~64KB (128 tokens × 128 dims). The dedup optimization saves GPU compute but then undoes the memory savings by cloning results back to full count. Consider using `Arc<Array2<f32>>` to share the allocation.

---

### Dead Code / Unused Code


### Correctness

**11. Pipeline error propagation may mask root causes**
`colgrep/src/index/mod.rs` ~lines 525–539 — Stages are joined sequentially. If the index stage errors first, the encode and pool stages continue processing until their sends fail, producing generic "Failed to send..." errors. The first stage joined (tokenize) reports that generic error rather than the real root cause from the index/metadata stage.

**Recommendation:** Consider collecting all stage results and reporting the first meaningful error, or using a shared `Arc<AtomicBool>` error flag so stages can abort early while preserving the original error.

---

### Quality / Minor

**15. `dynamic_batch` defaults to `true` in `ColbertBuilder::new`** (`next-plaid-onnx/src/lib.rs` ~line 814) — This is a behavioral change for library consumers who call `Colbert::builder(...)`. Document the change on the `with_dynamic_batch` builder method, or default to `false` for safety.

---

### Test Changes

The 20 parser test files are **mechanically correct** — every expected string consistently updated to place `File:` before `Code:`, matching the production change in `embed.rs`. No test coverage was lost. The four untouched test files (Julia, SQL, Zig, recursion guard) are correctly left alone.

---

### Summary

The most impactful items to address:
1. **Unbounded channels (#1)** — easy fix, high impact on memory for large repos
2. **CodeUnit clone during sort (#2)** — significant memory doubling
3. **Parallel compress_into_codes (#3)** — memory budget is per-batch, not total
4. **Dead code (#7, #8, #9, #10)** — straightforward cleanup
5. **Error masking (#11)** — users will see confusing "failed to send" instead of real errors

*AI generated*

cepera-ang:
> 1. Unbounded channels in the pipeline can cause memory blowup

idk, how much this is actually a concern, tbh. I made them unbounded on purpose -- memory demands seems to be fairly quaint compared to all other work and allowing first stages to run all the way means they unlikely to stall later when there is a lot of work being done on CPU with processing later stages. Basically CPU is underutilized in the beginning, before the GPU work and higher and higher utilized later, so it probably worthy trade off to a bit of memory for CPU. 

> 4. initial_create path accumulates all EncodedIndexChunks in memory before writing

That was an experiment to eliminate IO from the equation, should fine to bring back the incremental file updates (given they will be async/pipelined)

raphaelsty:
Closing this MR as MR #75 which is a copy paste of your MR has been merged, amazing work @cepera-ang, you rock, you are co-author on the MR #75 😊

PR #80, opened 2026-04-14
fix auto runtime-aware defaults before ORT init

RBozydar:
## Summary

Fix auto-mode indexing/search so runtime-aware `parallel` and `batch-size` defaults are resolved **after** ONNX Runtime / cuDNN initialization, instead of being frozen to CPU defaults too early.

## Problem

`colgrep` currently has two different concepts mixed together:

- "configured override from config/CLI"
- "effective value right now"

`Config::get_parallel_sessions()` and `Config::get_batch_size()` return effective values immediately. In auto mode, the command layer was calling those getters **before** ORT/cuDNN had been initialized.

On CUDA builds, that means `is_cudnn_available()` is still false at that point, so auto mode gets concretized to CPU defaults too early. In practice this can freeze values
like:
- `parallel = 16`
- `batch-size = 1`
into GPU indexing runs that should have stayed runtime-resolved.
That was causing bogus GPU fallback / poor GPU behavior in normal `colgrep init -y` runs even when forced GPU worked.

## Fix

This PR separates those two concepts more explicitly:

- Keep `get_parallel_sessions()` / `get_batch_size()` as "effective value now" getters
- Add explicit `configured_parallel_sessions()` / `configured_batch_size()` accessors for the late-resolution path
- Use those configured override accessors from `init` / `search`, so `None` stays `None` until `IndexBuilder` resolves defaults after ORT init
- Normalize malformed legacy `0` values through the configured override accessors
- Reuse a single CPU parallel-session fallback helper so the CPU fallback is consistent
- Update `colgrep settings` output to show `auto (runtime-resolved)` instead of eagerly displaying misleading concrete values

## Why this approach

I did **not** change the semantics of the existing `get_*()` getters, because they are still useful for places that genuinely want the effective value immediately.

The bug was the command layer using the wrong API for a runtime-sensitive path. This PR makes that distinction explicit instead of relying on raw field access or changing
getter behavior globally.

## Validation

- `cargo fmt --all`
- `cargo clippy -p colgrep --all-targets -- -D warnings`
- `cargo test -p colgrep`

Generated with Codex gpt-5.4 xhigh, reviewed by human

Also verified against a real repo with sourced runtime env:

- normal auto-mode `colgrep init -y repo` completed successfully without the bogus GPU fallback behavior

Generated by Codex 5.4 xhigh, reviewed by human.

### DIFF
--- a/colgrep/src/commands/config.rs
+++ b/colgrep/src/commands/config.rs
@@ -1,9 +1,23 @@
 use anyhow::Result;
 
 use colgrep::{
-    ensure_model, ensure_onnx_runtime, get_colgrep_data_dir, Config, DEFAULT_BATCH_SIZE,
-    DEFAULT_MAX_RECURSION_DEPTH, DEFAULT_MODEL, DEFAULT_POOL_FACTOR,
+    ensure_model, ensure_onnx_runtime, get_colgrep_data_dir, Config, DEFAULT_MAX_RECURSION_DEPTH,
+    DEFAULT_MODEL, DEFAULT_POOL_FACTOR,
 };
+
+fn format_parallel_setting(config: &Config) -> String {
+    config
+        .configured_parallel_sessions()
+        .map(|sessions| sessions.to_string())
+        .unwrap_or_else(|| "auto (runtime-resolved)".to_string())
+}
+
+fn format_batch_size_setting(config: &Config) -> String {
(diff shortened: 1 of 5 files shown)


raphaelsty:
Thank you for this MR, merging ☺️

PR #96, opened 2026-05-05
Consolidate agent install flags into --install-agent / --uninstall-agent, add Droid support

uepoch:
### Unified `--install-agent` / `--uninstall-agent` flags
All agent install and uninstall flags now go through two value-accepting flags backed by a `clap::ValueEnum`. Invalid agent names are caught at parse time with a clear error listing valid options.
### Droid (Factory) integration
Installs colgrep as a proper Factory skill at `~/.factory/skills/colgrep/SKILL.md` with YAML frontmatter, following Droid's skill discovery model. The Droid auto-discovers and invokes the skill when code search is needed, and users can also trigger it with `/colgrep`. Uninstall removes the skill directory entirely.
### Backward compatibility
`--install-opencode`, `--install-claude-code`, `--install-codex`, `--install-hermes` and their uninstall counterparts all continue to work but are hidden from `--help`.
If required, I can clear the old code, but I wanted to avoid breaking existing flows


uepoch:
Not sure if I can fix the `cc` linkage issue due to missing blas library on ubuntu. Let me know if this reappears

PR #105, opened 2026-05-22
Emit every matching line in regex mode (parity with grep)

raphaelsty:
## Summary

Compact regex mode (`colgrep -e <pattern>`) was silently dropping matches when multiple chunks of the same file collapsed into a single `SearchResult`. This PR makes regex emission scan the actual file from disk so output matches `grep` line-for-line, while leaving semantic-only output untouched.

## Root cause

`collapse_by_file` (src/index/mod.rs:3605) merges multiple matching chunks of the same file into one leader `SearchResult`, expanding `unit.line`/`unit.end_line` to span the whole region but **only keeping the leader chunk's `unit.code`**. Compact regex mode then scanned that one chunk's text via `find_matches_in_unit`, so matches living in the merged-in chunks were lost.

Concrete example on a 291-file repo, searching for `run_pipeline` (whole-word):

| File | grep hits | colgrep before | colgrep after |
|---|---|---|---|
| `run.py` | 4 (l. 64, 299, 339, 365) | 1 (l. 6 - a docstring) | 4 |
| `sources/indexer_daemon.py` | 7 | 1 | 7 |
| `sources/utils/index_health.py` | 5 | 1 | 5 |
| `scripts/reindex_broken.py` | 4 | 1 | 4 |
| ... | | | |
| **Total** | **38** | **19** | **38** |

## Change

Add `PatternMatcher::find_matches_in_file` that reads the file from disk and returns every `(line, content)` match (with the same literal-fallback semantics as `find_matches_in_unit`). Compact regex mode now:

1. Iterates `results` in score-descending order (already sorted upstream).
2. For each unique file, scans the file once and collects all matches.
3. Prints `file:line:content`, capped at `-k` total printed lines.

Semantic-only output path is unchanged: still emits `path:start-end` per chunk.

## Verified

On a 291-file mixed Python/Rust/JS repo (parity with `grep`):

- `-e run_pipeline -w`: **19 -> 38** matches (matches `grep`)
- `-e 'async fn' --include=*.rs`: **208** vs `grep`'s 208
- `-e 'Result<' --include=*.rs`: **103** vs `grep`'s 103
- `-e TODO -F` (case-insensitive): **10** vs `grep -i`'s 10
- Hybrid `-e 'fn .*auth' 'authentication'` still ranks by colgrep score
- Semantic-only output unchanged (`path:start-end` per chunk)
- 537/537 unit tests pass

## Test plan
- [x] Build with `cargo build --release`
- [x] `cargo test --release --lib` -> 537 passed
- [x] Manual stress test (counts above)
- [ ] Reviewer: run `colgrep -e <pattern>` on your own repo and confirm line counts match `grep -E`

### DIFF
--- a/colgrep/src/commands/search.rs
+++ b/colgrep/src/commands/search.rs
@@ -126,6 +126,48 @@ impl PatternMatcher {
             })
             .collect()
     }
+
+    /// Scan a file from disk for matching lines. Used by compact regex mode
+    /// so we don't miss matches in chunks that were merged into a leader
+    /// by `collapse_by_file` (the leader's `unit.code` only carries one
+    /// chunk's text). Returns 1-indexed `(line_number, line_content)`
+    /// pairs in source order.
+    fn find_matches_in_file(&self, file: &Path) -> Vec<(usize, String)> {
+        let content = match std::fs::read_to_string(file) {
+            Ok(c) => c,
+            Err(_) => return Vec::new(),
+        };
+
+        let scan = |needle: &dyn Fn(&str) -> bool| -> Vec<(usize, String)> {
+            content
+                .lines()
+                .enumerate()
+                .filter_map(|(i, line)| {
+                    if needle(line) {
+                        Some((i + 1, line.to_string()))
+                    } else {
+                        None
+                    }
+                })
+                .collect()
+        };
+
+        match self {
+            PatternMatcher::Regex(re) => {
+                let matches = scan(&|line| re.is_match(line));
+                if !matches.is_empty() {
+                    return matches;
+                }
+                // Literal fallback when the regex (often regex metacharacters
+                // a user meant literally) finds nothing.
+                let pattern_lower = re.as_str().to_lowercase();
+                scan(&|line| line.to_lowercase().contains(&pattern_lower))
+            }
+            PatternMatcher::Literal(pattern_lower) => {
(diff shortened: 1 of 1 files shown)


raphaelsty:
## Follow-up: line anchors (`^` / `$`) now work per-line

Added a second commit (8ea7d35) after running a regex-parity battery against grep on a 291-file repo. 18/19 patterns matched exactly out of the gate; two anchored patterns silently dropped matches because the SQLite `REGEXP` filter compiled the regex in single-line mode and matched it against the multi-line chunk text.

Forcing `(?m)` so anchors fire on every line inside a chunk restored parity:

| Pattern | Before | After | grep |
|---|---|---|---|
| `^use ` | 9 | **228** | 228 |
| `;$` | 3138 | **3517** | 3517 |
| `^pub fn` | 0 | **21** | 21 |
| `^impl.*for` | partial | **9** | 9 |
| `\.$` | partial | **3235** | 3235 |

Full battery (19 cases) now passes: dot/star/plus quantifiers, alternation, anchors, char classes (positive + negative), `\d`/`\w` shorthand, non-cap groups, group quantifiers, word boundaries, URL-ish patterns, unicode (`→|⇒|—`). `-F` literal mode and `-w` whole-word also pass, including `-w '^pub'` (anchor + word boundary together). Invalid regex / unsupported lookahead handled gracefully (no crash, empty result).

PR #106, opened 2026-05-22
Regex mode: per-line anchors + -k as document cap

raphaelsty:
Two follow-ups to #105 (whose first commit landed, but the second - the anchor fix - did not).

## 1. `(?m)` anchor fix (re-include of the commit dropped from #105)

`filter_by_text_pattern_with_options` compiled the regex against the multi-line \`code\` column with default (single-line) anchor semantics. So `^use ` only matched chunks whose *first byte* started with `use `, dropping every other line of a use-block. Prepending `(?m)` so anchors fire on every line restores grep parity:

| Pattern | Before | After (= grep) |
|---|---|---|
| `^use ` | 9 | **228** |
| `;$` | 3138 | **3517** |
| `^pub fn` | 0 | **21** |
| `^impl.*for` | partial | **9** |
| `\.$` | partial | **3235** |

## 2. `-k` as document cap; unbounded when omitted

In regex mode, grep has no implicit cap. colgrep was still applying its default `-k 15` to both the result-document count *and* the printed-line count, so:

```
colgrep -e foo            # silently truncated at 15 printed lines
colgrep -e foo -k 5       # truncated printed lines, not documents
```

New semantics:

```
colgrep -e foo            # every matching line in every matching document
colgrep -e foo -k N       # top N scored documents, every match inside those
colgrep "concept"         # unchanged: default -k 15, path:start-end format
colgrep "concept" -k 5    # unchanged
```

Threaded a `top_k_explicit: bool` (= `cli.top_k.is_some()`) into `cmd_search`. When `-e` is set and `-k` is not, the post-aggregation `.take(top_k)` becomes effectively unbounded (`usize::MAX / 1024`, large enough to be a non-cap and small enough that the upstream `top_k * 4` / `top_k * 20` multipliers can't overflow). Compact regex emission no longer caps printed lines either.

Semantic-only output path is untouched.

## Verified on the same 291-file repo

| Scenario | Result |
|---|---|
| `-e run_pipeline -w` (no -k) | 38 lines, exact grep parity |
| `-e run_pipeline -w -k 2` | 6 lines from 2 documents, footer prints |
| `-e run_pipeline -w -k 1` | 2 lines from 1 document, footer prints |
| Semantic-only (no -e, no -k) | 15 chunks (unchanged default) |
| Semantic-only `-k 5` | 5 chunks (unchanged) |
| 19-pattern regex battery (anchors, alternation, quantifiers, char classes, shorthand, unicode...) | all match grep exactly |
| `-F` literal, `-w` whole-word | parity |
| Invalid regex / unsupported lookahead | handled gracefully |
| 537/537 unit tests | pass |

## Test plan
- [x] `cargo build --release`
- [x] `cargo test --release --lib` -> 537 passed
- [x] Manual regex battery against grep (parity table above)
- [x] Pre-commit hooks (fmt + clippy + lib tests) green
- [ ] Reviewer: run `colgrep -e <pattern>` without `-k` on your own repo and confirm line count matches `grep -E`


raphaelsty:
## Follow-up: fixes the 3 escape bugs + adds lookaround / backref / case-sensitive

Added two more commits after the stress test in the previous comment caught real bugs and design limits.

### Commit e5dcafe — \`\{\` \`\}\` \`\+\` \`\?\` are now literal in ERE mode

Three pre-existing escape bugs in the BRE→ERE preprocessor:

| Pattern | Before | After (= grep) |
|---|---|---|
| `\+` (literal `+`, e.g. `a\+=`) | 0 | **218** |
| `\?` (literal `?`) | 0 | **546** |
| `\{` literal open brace | 0 | **4783** |
| `\}` literal close brace | 0 | **4782** |
| `fn.*?\{` (lazy + literal `{`) | 0 | **274** |
| `^pub fn .*\{$` | 0 | **18** |

Root causes:
- `escape_literal_braces` pushed the `\` then mangled the `{` into `[{]`, producing the unmatchable string `\[{]`. Now recognises `\{` / `\}` and emits `[{]` / `[}]` directly.
- `bre_to_ere` stripped the backslash from `\+` and `\?` unconditionally, turning ERE-literal `\+` into the one-or-more quantifier. Now kept escaped — matching GNU grep `-E` behaviour.

Balanced `\{n,m\}` still converts to `{n,m}` for BRE-style writers. Unit test `test_bre_to_ere_quantifiers` updated to assert the new (correct) semantics.

### Commit 692c7d0 — Lookaround, backreferences, and `-s` / `--case-sensitive`

Swap `regex` -> `fancy-regex` everywhere user-supplied patterns are matched (the SQLite REGEXP UDF in `next-plaid` and the per-line `PatternMatcher` in `colgrep`):

- Lookahead `foo(?=bar)` / negative `foo(?!bar)`
- Lookbehind `(?<=foo)bar` / negative `(?<!foo)bar`
- Backreferences `(\w+)\s*=\s*\1`

`fancy-regex` defers to the regex crate for patterns it can handle (i.e. every previously-working pattern) and only falls back to its NFA when a feature `regex` cannot express appears. The default `backtrack_limit` (1M) caps catastrophic patterns; `is_match`'s `Result<bool>` is treated as "no match" on backtrack-limit-exceeded so one pathological line cannot abort a whole scan.

Case-sensitivity is now controllable end-to-end via `-s` / `--case-sensitive`. Default stays case-insensitive (colgrep's historical behaviour); passing `-s` drops the `(?i)` inline flag in both the SQLite filter and the per-line matcher.

`PatternMatcher` is now a struct over `PatternKind { Regex, Literal }` instead of a bare enum, so the case flag lives next to the compiled pattern and line-match logic has a single source of truth.

Verified on the same 291-file repo:

| Feature | Matches |
|---|---|
| `fn (?=\w+_)` (lookahead) | **342** (was 0) |
| `(?<=async )fn` (lookbehind) | **211** (was 0) |
| `(\w+)\s*=\s*\1` (backref) | matches (was 0) |
| `-s STRUCT` / `-s struct` / default `STRUCT` | 0 / 189 / 189 (= grep / grep / grep -i) |
| 26/26 prior regex-parity battery | all pass |
| 537/537 unit tests | pass |

Architecture preserved: SQLite REGEXP UDF still does the chunk-level filter; per-line matcher still scans the file. Only the engine underneath changed.

PR #114, opened 2026-05-31
Add a Nix flake with packages for colgrep and a devshell

fu5ha:
Adds a basic `flake.nix` which provides:

- `colgrep` and `colgrep-cuda` packages
- A dev shell with the necessary environment to build colgrep
- A library function to build special packages set (only colgrep for now) with a specific cuda capabilities set, using the existing nix cuda pattern

This can be used by developers to quickly obtain a development environment using Nix, and can be used like so to add a specific colgrep package to a flake-based Nix environment like NixOS, home-manager, etc like so:

```nix
inputs = {
   // ...
   next-plaid = {
       url = "github:lightonai/next-plaid";
       inputs.nixpkgs.follows = "nixpkgs";
   };
};

outputs = inputs@{ self, nixpkgs, ... }:
let
    nextplaid-pkgs = (input.next-plaid.lib.mkPackagesWithCudaCapabilities [ "7.5" ]).${system};
in
{
    // ...
    // or wherever else packages are specified in one's config
    home.packages = [
        // ...
        nextplaid-pkgs.colgrep
    ];
};
```

It currently uses a semi-ugly wrapper style where we just inject environment variables. Ideally, the Rust packages that pull in the libs that need this wrapping (onnx-runtime, cuda libs) could be modified to allow link-time dynamic linking for these, which would then allow Nix's RPATH patching to let them link against the location of the libs in the Nix store without the need for wrapping with env vars. But, this isn't necessarily the end of the world for now.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1779560665,
+        "narHash": "sha256-tpyBcxPpcQb8ukyNF7DoCwfSY3VPsxHoYwj00Cayv5o=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "64c08a7ca051951c8eae34e3e3cb1e202fe36786",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-unstable",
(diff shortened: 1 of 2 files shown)


raphaelsty:
Hi @fu5ha, thank you for the MR ! Merging it right now 😊

PR #119, opened 2026-06-04
feat(onnx): add DirectML/MIGraphX/CoreML execution provider support

noctrex:
Extend colgrep's ONNX Runtime integration beyond CUDA so the same code path can drive non-NVIDIA GPUs via DirectML (Windows), AMD GPUs via MIGraphX, and Apple Silicon via CoreML, selected by cargo feature. The active backend is now also surfaced in the user-facing Model line.

next-plaid-onnx/src/lib.rs
- Add impl Display for ExecutionProvider with short canonical labels (CPU, CUDA, DirectML, CoreML, MIGraphX, TensorRT, auto) matching the tokens already used in onnx_runtime.rs download messages.
- Add test_execution_provider_display alongside the existing EP tests.

colgrep/src/index/mod.rs
- New cfg-gated IndexBuilder branch for directml/migraphx/coreml features that initializes the runtime and picks the matching ExecutionProvider. Falls through to the existing CPU branch when none of those features are enabled.
- Searcher::Auto (two sites) extends its fallback chain past CoreML to also consider DirectML and MIGraphX before Cpu.
- The Model: eprintln! now appends the selected backend in parentheses, e.g. "Model: lightonai/LateOn-Code (DirectML)". Provider is resolved at runtime so a CUDA build that falls back to CPU honestly prints CPU.

colgrep/src/onnx_runtime.rs
- Drop the USE_GPU bool constant in favor of per-feature eprintln!s so each provider gets an accurate label at download time.
- Route DirectML into the existing "gpu" cache subdir (shared with CUDA) and keep CPU in "cpu".
- Add a NuGet download path for Microsoft.ML.OnnxRuntime.DirectML on win-x64, since the GitHub release artifacts do not ship DirectML. Other platforms/configs keep using GitHub releases unchanged.

Note: MIGraphX and CoreML are wired into provider selection but their runtime download still falls through to the CPU package; users on those platforms are expected to supply the matching ORT build out-of-band for now. DirectML is the only new provider with full auto-download support.

Verified:
- cargo test -p next-plaid-onnx execution_provider (6 passed)
- cargo build --release -p colgrep --features directml on Windows x86_64
- cargo build --release -p colgrep (default features)
- Runtime: colgrep prints "Model: lightonai/LateOn-Code (DirectML)" with --features directml

Disclaimer: This commit was authored with assistance from an LLM.

### DIFF
--- a/colgrep/src/index/mod.rs
+++ b/colgrep/src/index/mod.rs
@@ -999,7 +999,12 @@ impl IndexBuilder {
                     }
                 }
             };
-            #[cfg(not(any(feature = "cuda", feature = "directml", feature = "migraphx", feature = "coreml")))]
+            #[cfg(not(any(
+                feature = "cuda",
+                feature = "directml",
+                feature = "migraphx",
+                feature = "coreml"
+            )))]
             let (num_sessions, execution_provider) = {
                 let _ = num_units;
 
--- a/colgrep/src/index/mod.rs
+++ b/colgrep/src/index/mod.rs
@@ -999,11 +999,10 @@ impl IndexBuilder {
                     }
                 }
             };
-            #[cfg(not(feature = "cuda"))]
+            #[cfg(not(any(feature = "cuda", feature = "directml", feature = "migraphx", feature = "coreml")))]
             let (num_sessions, execution_provider) = {
-                let _ = num_units; // Silence unused warning when cuda feature is disabled
+                let _ = num_units;
 
-                // Initialize ONNX Runtime (CPU-only build)
                 crate::onnx_runtime::ensure_onnx_runtime()
                     .context("Failed to initialize ONNX Runtime")?;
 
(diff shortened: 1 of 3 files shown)


aussetg:
Did you bench each backend ? In my own tests, Migraphx is **slow** ( see #122 ). But I can't test the rest.

noctrex:
@aussetg I am on Windows 11, with a AMD Ryzen 5800X3D CPU and a AMD RX 7900XTX GPU.
I just tried to index this very own codebase using the model LateOn-Code, using the DirectML backend to utilize the GPU.
Only this backend is available in Windows for AMD.

DirectML: 70,87 seconds
CPU: 643,25 seconds

Without this PR, it would not use DirectML at all and was always using only CPU.

raphaelsty (commented):
Amazing MR thank you 😊

PR #124, opened 2026-06-08
feat(colgrep): honor XDG_DATA_HOME on macOS

### LINKED ISSUE #123: XDG not honored on macOS
Hello!

Thanks for ColGrep, amazing tool.

I noticed from the docs that ColGrep supports XDG variables. However in practice ColGrep uses `dirs` to get the user's directories. This is great in principle, but the issue is that the `dirs` author is quite opinionated about XDG on macOS: he thinks the XDG env vars should not be honored even if they are defined (https://codeberg.org/dirs/dirs-rs/issues/56). This has led to `dirs` consumers being sometimes puzzled, and also fragmentation (see "xdg" and "dirs" on crates.io). For the record, Apple neither prevents the use of XDG, nor recommends against using it. 

Would you consider a few lines of code to honor XDG variables for the index location? (no change in Cargo.toml). I'm happy to open a PR if you're interested.

Thanks!

graelo:
Hi,

This is a follow-up on Issue #123 which I opened because the `dirs` crate intentionally ignores XDG env vars on macOS (see codeberg.org/dirs/dirs-rs/issues/56), so colgrep's index, marketplace, and uninstall paths landed under `~/Library/Application Support` regardless of `XDG_DATA_HOME`. As I mentioned in the issue, _preventing_ the use of XDG goes way beyond anything Apple mentions, let alone recommends, and as the `README` and env-var table already advertised `XDG_DATA_HOME` as supported, I reached out!

In practice, I added a small helper in `index::paths` that returns the XDG value when set and falls back to `dirs::data_dir()` otherwise, and simply route the three `dirs::data_dir()` call sites through it so the index location, the Claude Code marketplace dir, and the uninstall sweep all agree.

Closes #123

Thanks!

PS: tests pass (with cargo nextest run), and I get the expected behavior locally:

```console
ls ~/.local/share/colgrep/indices/
tmux-copyrat-5171936c
```

### DIFF
--- a/colgrep/src/index/paths.rs
+++ b/colgrep/src/index/paths.rs
@@ -62,6 +62,19 @@ impl ProjectMetadata {
     }
 }
 
+/// Base data directory, honoring `XDG_DATA_HOME` on all platforms (including
+/// macOS, where `dirs::data_dir()` ignores it by design — see
+/// <https://codeberg.org/dirs/dirs-rs/issues/56>).
+/// Falls back to `dirs::data_dir()` (platform default) when unset or empty.
+pub fn xdg_data_home_or_default() -> Result<PathBuf> {
+    if let Ok(xdg) = std::env::var("XDG_DATA_HOME") {
+        if !xdg.is_empty() {
+            return Ok(PathBuf::from(xdg));
+        }
+    }
+    dirs::data_dir().context("Could not determine data directory")
+}
+
 /// Get the base colgrep data directory (XDG_DATA_HOME/colgrep or platform equivalent).
 ///
 /// Honours the `COLGREP_DATA_DIR` env var so concurrent benchmark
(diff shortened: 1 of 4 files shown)


colgrep/src/index/paths.rs
```diff
@@ -62,6 +62,19 @@ impl ProjectMetadata {
     }
 }
 
+/// Base data directory, honoring `XDG_DATA_HOME` on all platforms (including
+/// macOS, where `dirs::data_dir()` ignores it by design — see
+/// <https://codeberg.org/dirs/dirs-rs/issues/56>).
+/// Falls back to `dirs::data_dir()` (platform default) when unset or empty.
```
graelo, line 68: I could go easier on this comment with just 

```rust
/// Base data directory, honoring `XDG_DATA_HOME` on all platforms (including
/// macOS).
```

You tell me.

raphaelsty:
Merging the MR :)

PR #125, opened 2026-06-09
Fix SQLite metadata contention during concurrent indexing and search

DeeJayPee:
## Summary

This PR fixes a production hang where concurrent metadata/search traffic could block while indexing or delete/update jobs were active. The issue was first observed during update_with_encoding, where /health and query endpoints could appear stuck, but the root cause was in the shared SQLite metadata/FTS access layer and could affect other endpoints too.

SQLite read paths were repeatedly opening/closing metadata databases and participating in connection setup behavior while write jobs were running. Under load, many requests piled up around SQLite connection open/close and blocked the API.

## Changes

- Split metadata SQLite access into read-only and write connection paths.
- Use cached read-only SQLite connections for metadata and FTS query paths.
- Keep WAL configuration on write/create paths only.
- Preserve snapshot-read behavior so search can continue while indexing writes are active.
- Avoid nested metadata DB reopen behavior in the fixed-schema update path.
- Replace temp-table based large read subsets with chunked read-only queries.
- Apply the same read/write split to text search metadata access.

## Validation

- `cargo check -p next-plaid` passed.
- `cargo fmt --package next-plaid` run.
- `git diff --check` passed.
- Focused test command was attempted, but currently blocked by an unrelated duplicate test name in `next-plaid/src/update.rs`.

## Notes

This does not change public API, CLI, request/response formats, or systemd configuration. The intended behavior change is that search, metadata, and health endpoints remain responsive while indexing is running.


raphaelsty:
Thank you for this MR, I was fixing the CI sorry to interupt you :)

raphaelsty:
LGTM, ready to merge

DeeJayPee:
Cool, thanks !
FYI, the fix has been battletested all night long on a production env.

PR #126, opened 2026-06-10
Remove redundant per-search and per-release work from colgrep indexing

voxmenthe:
When used on large repos, colgrep pays substantial costs on every search and update that buy nothing: every query re-reads and re-hashes every source file before returning results, every routine release discards every index on the machine, and every incremental update rewrites the index once per changed file. This has been annoying me, and this PR removes some of that redundant work.

Each of my commits preserves behavior. Search results, ranking, and the on-disk index layout are all unchanged.

## Commits

1. **Skip content hashing for files with unchanged mtime.** Restores the stat-only fast path from d0423d9 that was lost in the d76cb4a pipeline rewrite; without it every search reads and xxh3-hashes the entire repo. Also stops running orphan cleanup (a `SELECT DISTINCT` over every indexed file plus one stat each) on every search — it now runs only when the plan contains deletions, or every 50th search.
2. **Rebuild indexes on format changes, not on every CLI release.** `IndexState` now persists an `index_format_version`; rebuilds happen only when that constant is deliberately bumped, instead of whenever `cli_version` differs (which, with frequent patch releases via the auto-updating Homebrew tap, meant routinely re-embedding every project on the machine). Pre-existing indexes rebuild one final time; a missing `state.json` still takes the cheap reconstruct-from-DB path rather than a re-embed.
3. **Add `--no-update` to skip auto-indexing during search.** For agents issuing many queries in a row, the per-query change-detection scan is pure overhead. All three write paths are gated: the staleness check, the missing-index build, and the corrupt-index recovery rebuild.
4. **Flush incremental updates in 8K-doc batches.** The update path accumulated one index write per changed file; it now batches embeddings and flushes at a document threshold, collapsing N small rewrites into a few large ones.
5. **next-plaid: skip the full FTS5 rebuild when deleted docs are the ID-space tail.** Suffix deletes (the common truncate/re-add pattern) now delete the affected FTS rows directly instead of re-tokenizing the entire corpus; non-suffix deletes still rebuild because positional IDs renumber survivors.
6. **Size the default ONNX session count to the encoding workload.** Small updates no longer spin up the full default session pool to encode a handful of chunks.
7. **Load the user config once per command.** Each search re-parsed `config.json` up to seven-plus times across the `resolve_*` helpers and `search_single_path`; everything now reads one snapshot.
8. **Scan for a parent index only when the search path has none.** The parent-directory walk is skipped entirely when a local index already exists.

## Does it actually help?

For the headline claim (commit 1) I've added a committed measurement, ignored by default:

```
cargo test -p colgrep --release measure_update_plan -- --ignored --nocapture
```

On an Apple-silicon laptop with a warm page cache:

```
update plan, 5000 files x 200 KiB (976 MiB total):
  full hashing (pre-fix per-search cost): 522ms
  mtime fast path:                        131ms  (4.0x faster)
  read+hash component removed per search: 392ms (scales with repo bytes)
```

The 4.0x independently reproduces the "~4x query speedup" reported by d0423d9 when this fast path was first introduced. The measurement is conservative: hashing is timed against a warm page cache, and the eliminated component grows linearly with repo size (and worsens cold).

The other changes are pinned behaviorally: batched deletes assert exactly one index rewrite for N files (`test_delete_files_from_index_is_a_single_rewrite`), suffix deletes assert FTS survivor alignment without a rebuild (`test_delete_keeps_fts_aligned`), and the format-version change carries regression tests for both rebuild-once and reconstruct-instead-of-rebuild semantics.

## Known limitation (pre-existing, not addressed here)

colgrep's delete path renumbers metadata `_subset_` IDs without touching the FTS5 tables, so non-suffix deletes can leave FTS rowids misaligned with survivors (surfacing later as the non-fatal "FTS indexing failed" warning). The structural fix is stable document IDs, which I'm planning as a follow-up PR.

### DIFF
--- a/colgrep/src/cli.rs
+++ b/colgrep/src/cli.rs
@@ -430,6 +430,10 @@ pub struct Cli {
     #[arg(short = 'y', long = "yes")]
     pub auto_confirm: bool,
 
+    /// Skip the automatic index update and search the existing index as-is
+    #[arg(long = "no-update")]
+    pub no_update: bool,
+
     /// When to colorize output and syntax highlighting: auto, always, or never.
     /// `never` emits plain text with no ANSI escape sequences (useful for agents/pipes).
     #[arg(
@@ -549,6 +553,10 @@ pub enum Commands {
         /// Use strict batch-size batching instead of fixed dynamic GPU batching
         #[arg(long = "static-batch")]
         static_batch: bool,
+
+        /// Skip the automatic index update and search the existing index as-is
+        #[arg(long = "no-update")]
+        no_update: bool,
     },
 
     /// Show index status for a project
--- a/colgrep/src/commands/init.rs
+++ b/colgrep/src/commands/init.rs
@@ -36,10 +36,10 @@ pub fn cmd_init(path: &PathBuf, options: InitOptions<'_>) -> Result<()> {
         anyhow::bail!("Path is not a directory: {}", path.display());
     }
 
-    let model = resolve_model(options.cli_model);
-    let pool_factor = resolve_pool_factor(options.pool_factor, options.no_pool);
-
     let config = Config::load().unwrap_or_default();
+    let model = resolve_model(&config, options.cli_model);
+    let pool_factor = resolve_pool_factor(&config, options.pool_factor, options.no_pool);
+
     let quantized = !config.use_fp32();
     let (parallel_sessions, batch_size) =
         resolve_index_runtime_overrides(&config, options.batch_size);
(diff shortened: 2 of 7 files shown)


raphaelsty:
Hi @voxmenthe thank you for the MR, I need to review it carefully and then I'll merge it 😊

voxmenthe:
> Hi @voxmenthe thank you for the MR, I need to review it carefully and then I'll merge it 😊

Cool. I have another optimization PR ready when you're done with this one. Let me know if you have any questions thanks!

raphaelsty:
Thank you for your MR @voxmenthe, it has been merged within #132, I added you as a author of the MR :)

raphaelsty:
Please note I would love to welcome new MRs as this one which improve colgrep a lot :)

PR #134, opened 2026-06-17
fix(colgrep): eliminate unnecessary re-indexing on every search

vlasky:
## Summary

ColGREP's incremental indexing has several compounding issues that cause **every search to re-hash every tracked file**, even when nothing has changed. In our case (a Bitbucket directory with ~8,400 files across many repos and git worktrees), every search was reading and hashing all 8,400 files before returning results.

These fixes are small and focused (one file changed) but have a significant impact on usability. We'd appreciate a timely review as the current behavior makes colgrep impractical for large project directories.

## The problems

1. **Legacy states defeat the mtime fast path**: States written by <= 1.5.4 have `size: 0` (field didn't exist). The fast path requires `info.size == current_size`, which always fails when `info.size == 0`, causing every file to be content-hashed on every search.

2. **Format version 0 → 1 triggers a full rebuild**: Upgrading to 1.5.5 nukes the entire index and re-embeds all files from scratch, even though the on-disk format didn't actually change. For large projects this takes many minutes.

3. **Legacy mtime precision mismatch**: States written by <= 1.5.4 store mtimes in seconds, but the current fast path compares at nanosecond precision. The comparison always fails, defeating the fast path even after the format migration.

4. **Worktree seeding copies stale mtimes**: When seeding a new worktree from a sibling, the source's mtimes are persisted verbatim. Since git checkout stamps files with "now", every mtime mismatches on first search, triggering a full content-hash pass.

5. **Redundant `canonicalize()` syscalls**: `is_within_project_root` calls `realpath(3)` on every file during scan, despite `follow_links(false)` already preventing symlink escape. This adds ~2N unnecessary syscalls.

## The fixes (one per commit)

- **`perf(index): remove redundant canonicalize in scan_files`**: Skip `canonicalize()` when the relative path has no `..`; fall back to full validation only for the rare case where it does.

- **`fix(index): tolerate size == 0 in mtime fast path for legacy states`**: Treat `size == 0` as "unknown (legacy)" and use mtime-only comparison. This immediately restores the fast path for all pre-1.5.5 states.

- **`fix(index): migrate format version 0 → 1 in place instead of full rebuild`**: Bump the version in state.json without discarding the index. Future genuinely-incompatible format changes still trigger a full rebuild.

- **`fix(index): refresh file stats after worktree seeding`**: After copying the source state, stat each tracked file in the current worktree to stamp correct local mtimes/sizes before persisting. Also enhances the format 0 → 1 migration to stat all entries: fills in missing sizes, upgrades second-precision mtimes to nanoseconds (when the file hasn't been modified since last indexing), and purges entries for files that no longer exist on disk.

## Testing

- All 612 existing unit tests pass.
- Stress-tested against a 100-file git repo with worktrees: verified the fast path hits correctly, format migration upgrades mtimes/sizes and purges stale entries without rebuild, worktree seeding produces correct local mtimes, and modifications are still detected.
- Applied to our production index (~8,400 files): search now returns results immediately instead of re-hashing everything.

## Impact without these fixes

On a project with N tracked files, **every single search** performs N file reads + N hash computations before returning any results. With worktrees, this penalty is paid independently for each worktree on first use. On upgrade to 1.5.5, the entire index is additionally discarded and rebuilt from scratch.

### DIFF
--- a/colgrep/src/index/mod.rs
+++ b/colgrep/src/index/mod.rs
@@ -2678,31 +2678,20 @@ fn is_file_too_large(path: &Path) -> bool {
 /// The check is done by canonicalizing both paths and verifying
 /// the resolved path starts with the project root.
 fn is_within_project_root(project_root: &Path, relative_path: &Path) -> bool {
-    // Check for obvious path traversal patterns first (fast path)
     let path_str = relative_path.to_string_lossy();
-    if path_str.contains("..") {
-        // Could be a traversal attempt - do full canonicalization check
-        let full_path = project_root.join(relative_path);
-        match full_path.canonicalize() {
-            Ok(canonical) => {
-                // Canonicalize project root as well for accurate comparison
-                match project_root.canonicalize() {
-                    Ok(canonical_root) => canonical.starts_with(&canonical_root),
-                    Err(_) => false,
-                }
-            }
-            Err(_) => false, // If canonicalization fails, reject the path
-        }
-    } else {
-        // No ".." in path, but still verify the path doesn't escape via symlinks
-        let full_path = project_root.join(relative_path);
-        if !full_path.exists() {
-            return true; // Non-existent paths will be skipped later anyway
-        }
-        match (full_path.canonicalize(), project_root.canonicalize()) {
-            (Ok(canonical), Ok(canonical_root)) => canonical.starts_with(&canonical_root),
-            _ => false,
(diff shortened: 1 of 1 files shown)


raphaelsty:
Hi @vlasky, thank you for this MR, I merged it within #137 and apply few updates

PR #135, opened 2026-06-17
perf(index): reduce incremental delete from O(index_size) to O(IVF_size)

vlasky:
## Summary

- Replace full IVF rebuild (which read all chunk code files into a BTreeMap) with in-place IVF patching that filters deleted IDs and renumbers survivors using binary search
- Replace temp-table-copy re-sequencing in SQLite (which copied and reinserted the entire ~5GB METADATA table) with targeted range-based UPDATE statements on the integer PK
- Add `MmapIndex::update_append` to skip merged file regeneration during incremental updates (deferred to next search)

On a ~90K doc index with 12GB vector data and 5GB metadata, incremental updates of 1-3 files drop from ~60s to ~3s.

## Test plan

- [ ] `cargo test` - all 768 tests pass
- [ ] `cargo clippy --all-targets -- -D warnings` - clean
- [ ] `cargo fmt --all -- --check` - clean
- [ ] Manual: modify 1-3 files in a large indexed project, run `colgrep index`, verify completes in <5s
- [ ] Manual: search results after incremental update match expected content

PR #138, opened 2026-06-17
perf(index): reduce incremental delete from O(index_size) to O(IVF_size) (rebase of #135)

raphaelsty:
## Summary

Rebase of **#135** (by @vlasky) onto current `main` (post #137). No conflicts — the FTS5-in-delete code #135 builds on came from #132 (already in 1.5.5), so it merged cleanly.

The change reduces incremental-delete cost from O(index_size) to ~O(IVF_size):
- **In-place IVF patch** instead of a full rebuild that re-read every chunk's codes (filter deleted IDs, renumber survivors via binary search).
- **Range-based SQLite re-sequencing** (`UPDATE ... SET _subset_ = _subset_ - shift`) instead of copying the whole METADATA table through a temp table.
- **`MmapIndex::update_append`** to skip merged-file regeneration during incremental updates (deferred to the next search, which regenerates lazily).
- colgrep skips the intermediate FTS5 rebuild on deletes that are immediately followed by re-encoding (`delete_files_from_index_no_fts_rebuild`).

## Status / review notes (see PR discussion)
- Validated end-to-end: keyword (`-e`) search returns correct results after incremental edits/deletes (deleted content not surfaced, survivors resolve correctly).
- Open considerations being measured/assessed: raw-FTS rows are left stale until a delete-only update rebuilds them; range re-sequencing assumes in-range delete IDs; IVF patch preserves (does not dedup) existing per-centroid entries.

Co-authored with @vlasky (#135).

### DIFF
--- a/colgrep/src/index/mod.rs
+++ b/colgrep/src/index/mod.rs
@@ -86,6 +86,18 @@ fn delete_from_index_counted(ids: &[i64], index_path: &str) -> Result<usize> {
 /// surviving documents, so interleaving reads and deletes would invalidate the IDs that haven't
 /// been deleted yet. Returns the number of documents removed.
 fn delete_files_from_index(index_path: &str, files: &[PathBuf]) -> Result<usize> {
+    delete_files_from_index_inner(index_path, files, true)
+}
+
+fn delete_files_from_index_no_fts_rebuild(index_path: &str, files: &[PathBuf]) -> Result<usize> {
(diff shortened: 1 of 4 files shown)


raphaelsty:
Superseded by #139 (merged into main as 8630af3), which includes this rebase plus the hardening fixes (id clamping, FTS5 pruning, lazy merged-file regen, sync/dedup tests, backward-compat). Closing as redundant.

PR #144, opened 2026-06-18
perf(filtering): split METADATA into thin + fat tables (v2 schema)

### LINKED ISSUE #136: perf(filtering): split METADATA into thin index + fat content table

vlasky:
## Summary

Splits the single METADATA table into a thin table (~50 bytes/row) and a fat METADATA_CONTENT table (large TEXT columns), linked by a stable monotonic `_content_id_`. Re-sequencing after deletes now touches only the thin table, making performance position-independent.

**Benchmarks** (20K rows, 27KB each, delete near position 0):
- v0 (INTEGER PRIMARY KEY): 940ms
- v1 (#141, demoted PK): 211ms
- **v2 (this PR, thin table): 8ms** - 26x faster than v1, 117x faster than v0

## Design

- `METADATA` retains `_subset_`, `_content_id_` FK, and small filterable columns (file, name, line, unit_type, language, complexity, booleans)
- `METADATA_CONTENT` holds large TEXT columns (code, signature, docstring, parameters, calls, etc.) keyed by `_content_id_` INTEGER PRIMARY KEY
- `_content_id_` is stable and monotonic - never re-sequenced on delete
- `PRAGMA user_version = 2` marks the new layout
- Orphaned content rows are cleaned up on delete

## Backward compatibility

- v0/v1 read and write paths preserved unchanged - existing indices work with no migration
- `INDEX_FORMAT_VERSION` bumped to 2 in colgrep, triggering a full rebuild on upgrade (no in-place migration needed)
- Removed the v0->v1 in-place migration path; format mismatches now trigger full rebuild

## Changes

- `next-plaid/src/filtering.rs`: v2 `create`, `update_v2`, `delete_v2`, `get_v2`, `update_where_v2`, JOIN-based `where_condition`/`where_condition_regexp` for fat column queries
- `next-plaid/src/text_search.rs`: FTS `rebuild()`/`update_rows()` adapted to JOIN for v2
- `colgrep/src/index/state.rs`: `INDEX_FORMAT_VERSION: 1 -> 2`
- `colgrep/src/index/mod.rs`: simplified format mismatch handling to full rebuild

## Related

- Closes #136 (thin/fat split proposal)
- Builds on #141 (demoted PK - the v1 approach this supersedes for new indices)

## Test plan

- [x] 130 unit tests in next-plaid pass (including 6 new v2-specific tests)
- [x] 619 tests in colgrep pass
- [x] 54 tests in next-plaid-api pass
- [x] Integration tested on 3,722-file real-world index (search, incremental update, format-version-triggered rebuild)
- [x] Stress tested: delete at all positions, concurrent readers+writer, adversarial inputs (1MB blobs, Unicode, NUL bytes, SQL injection attempts, out-of-range IDs)

### DIFF
--- a/colgrep/src/index/mod.rs
+++ b/colgrep/src/index/mod.rs
@@ -1634,7 +1634,7 @@ impl IndexBuilder {
         // instead of rebuilding from scratch. No-op for non-worktree projects.
         self.maybe_seed_from_worktree(force);
 
-        let mut state = IndexState::load(&self.index_dir)?;
+        let state = IndexState::load(&self.index_dir)?;
         let index_dir = get_vector_index_path(&self.index_dir);
         let index_path = index_dir.to_str().unwrap();
         let index_exists = index_dir.join("metadata.json").exists();
@@ -1660,41 +1660,10 @@ impl IndexBuilder {
             return self.full_rebuild(languages);
         }
 
-        // Format version 0 → 1 did not change the on-disk index layout, only
-        // added version tracking and the FileInfo.size field (which defaults to
-        // 0 via serde). Migrate in place rather than discarding the entire index.
-        if format_mismatch && state.index_format_version == 0 {
-            state.index_format_version = INDEX_FORMAT_VERSION;
-            // Stat every still-present entry to fill in missing sizes and upgrade
-            // legacy second-precision mtimes, restoring the stat-only fast path
-            // without re-embedding. Entries whose file is gone are deliberately
-            // LEFT in the state: the incremental update below will see them
-            // missing from disk, mark them deleted, and remove them from every
-            // store (vector index + metadata + FTS5) via delete_files_from_index.
-            // Dropping them here would instead orphan their index rows — the
-            // deleted content would stay searchable until a later periodic cleanup.
-            for (path, info) in state.files.iter_mut() {
-                let full = self.project_root.join(path);
-                let Ok((mtime, size)) = file_stat(&full) else {
-                    continue; // gone on disk → handled as a deletion by the update
-                };
-                if info.size == 0 {
-                    info.size = size;
-                }
-                // Legacy states store mtime in seconds; current code uses
-                // nanoseconds. Upgrade precision when the file hasn't been
-                // modified (same second), otherwise leave the value stale so the
-                // hash pass below catches the real change.
-                if info.mtime < 10_000_000_000_000 {
-                    if info.mtime == mtime / 1_000_000_000 {
-                        info.mtime = mtime;
-                    }
-                } else if info.mtime == 0 {
-                    info.mtime = mtime;
-                }
-            }
-            state.save(&self.index_dir)?;
-        } else if format_mismatch {
+        // Format version 0/1 to 2 changes the metadata DB schema (thin/fat table
+        // split), requiring a full rebuild. Drop any in-progress resumable-build
+        // marker so we don't try to resume an index we're discarding.
+        if format_mismatch {
             let _ = std::fs::remove_file(self.index_dir.join(BUILDING_MARKER));
             return self.full_rebuild(languages);
         }
--- a/colgrep/src/index/state.rs
+++ b/colgrep/src/index/state.rs
@@ -12,7 +12,7 @@ use super::paths::get_state_path;
 /// metadata schema). Bump ONLY for incompatible changes: a mismatch discards
 /// the index and re-embeds the entire project on the next run. Routine CLI
 /// releases must NOT bump this.
-pub const INDEX_FORMAT_VERSION: u32 = 1;
+pub const INDEX_FORMAT_VERSION: u32 = 2;
 
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub struct IndexState {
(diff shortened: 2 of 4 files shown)


raphaelsty:
## Benchmark: v2 (this PR) vs main (v1) — wins & losses

Same `next_plaid::filtering` code compiled against each branch (schema differs internally). Corpus: **50K docs, ~24 KB `code`/row** (also 20K/100K for delete scaling). Delete measured with a **no-copy** harness (build fresh → time only the delete) to avoid WAL-checkpoint contamination; reads/build/size measured on the built index. CPU, release build.

### Delete — front delete of 30 units (the headline)
| index | v1 (main) | v2 (this PR) | speedup |
|---|---:|---:|---:|
| 20K | 408 ms | 38 ms | **10.7×** |
| 50K | 1152 ms | 101 ms | **11.5×** |
| 100K | 2283 ms | 209 ms | **10.9×** |

Consistent **~11×**. Note both remain **O(N)** for a front/early delete (linear in rows shifted) — v2 just has an ~11× smaller per-row constant (thin rows, no fat-record rewrite). So it's a large constant-factor win rather than truly position-independent; end/recent deletes are near-instant either way.

### Read / build / size (50K)
| operation | v1 (main) | v2 (this PR) | result |
|---|---:|---:|---|
| `where_condition` on a **thin** col (file/lang/unit_type/complexity) | 188.9 ms | **2.5 ms** | 🟢 **~75× faster** |
| `get(100 ids)` — search-result retrieval | 0.82 ms | 0.92 ms | ⚪ unchanged |
| DB size | 1332 MB | 1337 MB | ⚪ +0.4% |
| build / insert (50K) | 4252 ms | 4552 ms | 🟡 +7% (two-table insert) |
| `where_condition` on a **fat** col (code/signature…), JOIN | 199.8 ms | 233.9 ms | 🟡 +17% (JOIN) |

### Wins & losses
**Wins** 🟢
- **Delete ~11× faster** at all sizes.
- **Thin-column filtering ~75× faster** — not mentioned in the PR but arguably the biggest practical win: the filter scans 50-byte thin rows instead of 24 KB fat rows, which is exactly what colgrep's `--include` / language / `--code-only` / hybrid metadata filters hit.

**Neutral** ⚪ — search-result `get` retrieval and on-disk size are unchanged.

**Losses** 🟡 (both minor)
- **Fat-column filtering +17%** (JOIN) — only matters if filtering directly on `code`/`signature`/etc., which is rare (full-text goes via FTS).
- **Insert/build +7%** (writing two tables).

### Bottom line
Net strongly positive: big wins on delete (~11×) and thin-column filtering (~75×), no size penalty, search retrieval unaffected, against small well-contained losses. The two non-perf caveats are unchanged from review: front deletes are still O(N) (just ~11× cheaper, not truly position-independent), and the layout drops **forward-compat** (an old next-plaid-api reading a v2 index gets thin columns only) — mitigated for colgrep via `INDEX_FORMAT_VERSION` rebuild, but not for the API.

PR #149, opened 2026-07-02
feat(colgrep): first-class ops, build & API-schema formats (Terraform, Bash, PowerShell, Protobuf, GraphQL, Starlark, CMake, Groovy, INI)

raphaelsty:
## What

Extends colgrep's language coverage across the ops / build / API-schema surface. Every format below was previously either **skipped entirely** (unknown extension) or indexed as **one blob per file** (text pipeline). They now produce named, line-anchored semantic units via tree-sitter.

| Format | Extensions / filenames | Unit = | Example unit name |
| --- | --- | --- | --- |
| **Terraform / HCL** | `.tf`, `.tfvars`, `.hcl` | block | `resource "aws_instance" "web"` |
| **Shell** ⬆️ | `.sh`, `.bash`, `.zsh` | function | `deploy_app` |
| **PowerShell** ⬆️ | `.ps1`, `.psm1`, `.psd1` | function / class | `Deploy-App` |
| **Protobuf** | `.proto` | message / enum / service | `service Billing` |
| **GraphQL** | `.graphql`, `.gql` | type / query / mutation / fragment / … | `mutation CreateUser` |
| **Starlark / Bazel** | `.bzl`, `.star`, `BUILD`, `WORKSPACE`, `MODULE.bazel` | target / macro | `cc_library "mylib"` |
| **CMake** | `.cmake`, `CMakeLists.txt` | function / macro | `add_catch_test` |
| **Groovy** | `.groovy`, `.gradle`, `Jenkinsfile` | class / method / function | `bumpSemver` |
| **INI** | `.ini`, `.cfg`, `.properties`, systemd `.service`/`.timer`/`.socket` | section | `[database]` |

⬆️ = upgraded from text treatment. **Behavior change:** shell/PowerShell scripts now count as *code* for `--code-only` and rank in the code partition. Existing indexes pick this up on rebuild.

Plus zero-cost extension aliases onto existing grammars: `.pyi`→Python, `.mts`/`.cts`→TypeScript, `.cjs`→JavaScript, `.sbt`→Scala, `.rake`/`.gemspec`/`Rakefile`/`Gemfile`/`Vagrantfile`→Ruby, `.mk`→Makefile. The `.hcl` support also covers Packer, Nomad, Vault, Consul, and Waypoint configs.

## AST chunks vs raw files — measured, not assumed

We ran a retrieval A/B eval per format: identical corpora indexed twice (real extensions → AST units, `.txt` → whole-file documents), 47 queries with known target files, scored by MRR / hit@1.

- **Result: parity (MRR 1.0 both variants) on 8 of 9 formats, one AST miss on CMake.**
- The miss was diagnostic gold: the only mention of "catch2 / ctest" for a helper lived in the `#` comment *above* the function, which wasn't part of the unit. Fixed by attaching leading `#` doc comments to units for CMake, Starlark and Terraform (Shell already had it; regression test added). **After the fix: MRR 1.0 everywhere.**
- Verdict: AST treatment kept for all formats — equal retrieval quality, and it additionally gives named units, exact line ranges, and correct `--code-only` semantics. Raw would win only by accident of comment placement, which is now moot.

## Investigated and deliberately not included

- **nginx**: `tree-sitter-nginx` mis-parses `upstream` blocks and errors on IP literals (`server 10.0.0.1:8080`) — real configs would chunk garbage.
- **Dockerfile / CUE / SCSS grammars**: crates stuck on pre-0.25 tree-sitter ABI (Dockerfile stays on the text pipeline, which suits line-oriented files).
- **`.env`**: deliberately unindexed — secrets-leak vector.
- **YAML restructuring**: explicitly out of scope by decision; YAML keeps its current document treatment.

## Implementation notes

- All node kinds verified against real parse-tree dumps before wiring (no guessed grammar names).
- Starlark targets: only calls with a `name = "..."` **string** kwarg become units, so `glob()`, `load()`, and macro-forwarded `name = name` calls don't produce junk units.
- Heredoc/robustness invariants from the Terraform work (in-bounds line ranges, non-empty names, 100% non-empty-line coverage) are now a shared harness in `tests/common.rs`, applied to every new format.
- Earlier commits in this PR: Terraform support + stress suite, and a general `end_line` past-EOF clamp for unterminated constructs (all four extractors).

## Testing

- **458 parser tests pass** (8 new per-format test files, one per filetype per house convention + detection tests for every new extension/filename); clippy and fmt clean.
- End-to-end verified: indexing + semantic & regex search over the eval corpora (73 files, 9 formats); Terraform additionally stress-tested with a 131-file/~5,700-block adversarial corpus (huge files, 200-deep nesting, heredoc traps, unicode, malformed input).


raphaelsty:
Closed automatically by the head-branch rename (`feat/terraform-support` → `feat/ops-formats`). Continued in #150 with identical history — Terraform commits plus the ops/build/API-schema formats extension.

PR #151, opened 2026-07-03
perf(fts): key FTS5 by stable _content_id_ so deletes never rebuild

vlasky:
## Summary

Follow-up to #144. After the thin/fat split made metadata re-sequencing cheap, the dominant cost of an incremental update became the **FTS5 index**: any non-suffix delete forced a full O(corpus) `text_search::rebuild`, because FTS rows were keyed by `_subset_`, which `filtering::delete` re-sequences. On a 20K-unit index (234 MB metadata.db) the metadata delete costs 25 ms while the forced rebuild costs **835 ms**; at 50K×24KB scale it is multi-second, paid on every update that touches a non-tail file.

This PR keys the FTS by the stable, monotonic `_content_id_` that #144 introduced, as a **contentless** FTS5 table (`content=''`, `contentless_delete=1`; needs SQLite >= 3.43, the bundled build is 3.51):

- `filtering::delete` removes the deleted docs' FTS rows inside its own transaction; `_subset_` re-sequencing never touches FTS rowids, so **no rebuild ever happens again**.
- `METADATA_FTS_CONTENT` is gone for the new layout: no second copy of the corpus text. Inserts write one FTS row instead of a content row + FTS row.
- Searches JOIN the thin table (new index on `METADATA(_content_id_)`) to translate rowids back to `_subset_`. Public API and result semantics are unchanged.

## Benchmarks (`next-plaid/tests/fts_delete_bench.rs`, 20K docs x 4KB code, IdentifierAware, release)

| metric | before (subset-keyed) | after (content-id keyed) |
|---|---:|---:|
| middle delete of 30 units incl. FTS upkeep | ~860 ms (25 ms delete + 835 ms rebuild) | **25.4 ms** |
| metadata.db size | 234 MB | **136 MB (-42%)** |

The FTS upkeep inside the delete is unmeasurable next to the thin-table re-sequencing (25.39 ms vs 25.40 ms).

## Correctness fixes that fall out of the design

The legacy layout indexed the **pre-tokenized** text on insert but **raw** text on rebuild/`update_rows`:
- `IdentifierAware` sub-token matching (query `parse` hitting `parseRequest`) silently degraded after the first rebuild, i.e. after colgrep's first incremental update touching a non-tail file.
- FTS5's external-content `'delete'` command requires the exact indexed text; passing the raw content text for rows inserted with pre-tokenized text can corrupt the inverted index.

The new layout always indexes the tokenizer-prepared form, and deletes by rowid, so neither mismatch can occur.

## Backward compatibility / migration

- v0/v1 (non-split) metadata DBs keep the legacy subset-keyed FTS behaviour unchanged.
- Existing v2 indices with a legacy FTS keep working (reads, appends, suffix deletes) and are **migrated by `text_search::rebuild()`**, which is exactly the call the legacy layout needed after any non-suffix delete anyway. The migration therefore costs one already-scheduled rebuild, and afterwards deletes are O(deleted) forever. No `INDEX_FORMAT_VERSION` bump, no forced re-embedding.
- `MmapIndex::delete_with_options` and colgrep's `delete_files_from_index_inner` skip the delete/rebuild follow-up when the index is content-id keyed.
- Forward-compat caveat (an old binary reading a new index gets misaligned FTS ids until its own rebuild heals it to legacy) is the same stance #144 took for the split schema itself.

## Test plan

- [x] Full workspace suite: 877 tests pass (next-plaid 130 unit + 24 FTS integration incl. 6 new content-keyed tests and a legacy-to-content-id migration test; colgrep 621; next-plaid-api 54)
- [x] New integration tests: middle delete with no rebuild, repeated front deletes + append, IdentifierAware sub-tokens surviving deletes, `text_search::delete` before `filtering::delete`, legacy migration via `rebuild()`
- [x] Reproducible benchmark: `cargo test --release -p next-plaid --test fts_delete_bench -- --ignored --nocapture`
- [x] `cargo fmt` and `cargo clippy --workspace --all-targets` clean

### DIFF
--- a/colgrep/src/index/mod.rs
+++ b/colgrep/src/index/mod.rs
@@ -125,8 +125,11 @@ fn delete_files_from_index_inner(
     // Vector index (codes/residuals/IVF). Single batched rewrite (issue #116).
     delete_from_index_counted(&ids, index_path)?;
 
-    // Metadata and FTS5 must be deleted together and kept aligned. `filtering::delete`
-    // re-sequences the surviving `_subset_` IDs, so the FTS5 rows have to follow: a
+    // Metadata and FTS5 must be deleted together and kept aligned. On the current
+    // content-id keyed FTS layout the rowids are stable `_content_id_` values, so
+    // `filtering::delete` maintains the FTS itself and nothing more is needed here.
+    // On the legacy subset-keyed layout, `filtering::delete` re-sequences the
+    // surviving `_subset_` IDs, so the FTS5 rows have to follow: a
     // tail-only delete keeps every survivor's ID (drop just the deleted FTS rows,
     // O(deleted)); any other delete shifts survivor IDs and needs a full FTS5 rebuild.
     // This mirrors `next_plaid::MmapIndex::delete_with_options` — the standalone delete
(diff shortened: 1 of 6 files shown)


raphaelsty:
Thank you for this MR @vlasky, this is really cool, I'm reviewing it and merging asap :)

PR #153, opened 2026-07-07
fix(search): always check parent index regardless of cwd

vlasky:
## Summary

- Removes the `is_external_project_path` gate that skipped `find_parent_index` when the search path was not a subdirectory of cwd
- Fixes: searching an absolute path from an unrelated working directory failed to find the parent index, causing either a redundant index build or no results
- The `local_index_exists` check already short-circuits the common case; `find_parent_index` is cheap (one `read_dir` + one JSON parse per indexed project)

## Repro

```bash
# Index a project
colgrep init ~/projects/myrepo

# Search a subdirectory from an unrelated cwd
cd /tmp
colgrep "some query" ~/projects/myrepo/src/
```

Expected: uses existing index for `~/projects/myrepo`
Actual (before fix): fails to find the parent index because `/tmp` is not an ancestor of the search path

## Test plan

- [x] `cargo build -p colgrep` clean
- [x] `cargo clippy -p colgrep -- -D warnings` clean
- [x] `cargo test -p colgrep --lib` passes (662 tests)
- [ ] Manual: search a subdirectory of an indexed project from an unrelated cwd, confirm parent index is used

### DIFF
--- a/colgrep/src/commands/search.rs
+++ b/colgrep/src/commands/search.rs
@@ -410,16 +410,6 @@ fn display_path_with_cwd(path: &Path, cwd: Option<&Path>, use_relative: bool) ->
     normalized_path.display().to_string()
 }
 
-fn is_external_project_path(search_path: &Path, cwd: Option<&Path>) -> bool {
-    let Some(cwd) = cwd else {
-        return false;
-    };
-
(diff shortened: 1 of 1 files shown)


raphaelsty:
Amazing MR @vlasky you rock 🤘☺️

PR #155, opened 2026-07-07
feat: asymmetric binary quantization — int8 query × 1-bit docs with fused SIMD MaxSim kernels

joe32140:
## What

Optional binary document storage (`IndexConfig { binary: true }`): document
embeddings are stored as packed sign bits — **32× smaller than f32, 8× smaller
than the nbits=4 residual codec** — and scored asymmetrically against an
int8-quantized query with no float decompression, adapting mixedbread's
"int8 × binary" asymmetric quantization to multi-vector ColBERT MaxSim.

## How

Scoring uses the exact identity `q·s = 2·Σ_{s=+1} q − Σq` (2P − T), so every
kernel produces bit-identical integer scores — platform dispatch changes
speed, never ranking.

For byte-aligned dims (`dim % 8 == 0`, up to 256), fused doc-token-outer
kernels expand each doc token's sign bits once in registers, amortized over
all query tokens (monomorphized over the per-token group count so the hot
loops stay statically unrolled):
- **AVX-512 VNNI** (`vpdpbusd`), doc tokens in blocks of 4, vectorized max
- **AVX2 masked-SAD** (`psadbw` over biased u8 query codes)
- **NEON SDOT** on ±1 lanes (signed×signed, so `q·s` needs no 2P−T detour), blocks of 4, `vpaddq` reduction

Ragged dims and dims > 256 fall back to per-pair SIMD dots (NEON / AVX2 /
scalar). Dispatch
validates shapes and prepared-layout consistency — no UB reachable through
the public API. Incremental update is rejected for binary indexes at the
`update_index` root (regression-tested) so appends can't corrupt a sign-bit
store.

## Kernel microbenchmark (32 query × 80 doc tokens per doc)

Apple M4 (native NEON), across the dims real ColBERT models ship at:

| dim | fused NEON | per-pair fallback | speedup |
|---:|---:|---:|---:|
| 128 | 1.98 µs/doc | 13.3 | 6.7× |
| 96 | 1.54 | 10.1 | 6.5× |
| 64 | 1.16 | 6.8 | 5.9× |
| 48 | 0.94 | 5.4 | 5.8× |

Cross-platform (GitHub runners, dim=128, full test suite + bench per ISA):

| runner | kernel | fused | fallback | speedup |
|---|---|---:|---:|---:|
| windows x86 | AVX-512 VNNI | 3.23 µs | 16.8 | 5.2× |
| linux x86 ×2 | AVX2 masked-SAD | 4.53 µs | 20.5 | 4.5× |
| linux arm64 (Neoverse) | NEON SDOT | 4.69 µs | 40.2 | 8.6× |
| macos arm64 | NEON SDOT | 2.88 µs | 21.4 | 7.4× |

A full dim sweep (48/64/96/128/256) on the same five runners confirms the
fused kernel beats the per-pair fallback at **every dim on every ISA**
(2.8×–8.6×; e.g. NEON dim=256: 10.9 vs 77.9 µs/doc), and the dispatched
entry point always matches the forced kernel. Runner hardware is
heterogeneous — draws without AVX-512 VNNI land on the AVX2 path, so both
x86 tiers are exercised in practice.

## End-to-end results on real BEIR data (SciFact)

Full-pipeline profile on this index implementation: 5,183 docs / 1.19M doc
tokens / 300 judged queries, LateOn checkpoints (dim=128), Apple M4 with
Accelerate. NDCG@10 below is the deployed (PLAID-default ANN) regime. The
ceiling is an independent exhaustive numpy brute force (every query scored
against every document, no ANN, no candidate cap) — the deployed binary
NDCG matches that ceiling within rounding (e.g. LateOn-regularized: deployed
0.7513 vs exhaustive 0.7512).

| model | scheme | index (MB) | B/token | vs f32 | NDCG@10 | retention | mean ms/query | p95 ms | speedup |
|---|---|---:|---:|---:|---:|---:|---:|---:|---:|
| LateOn | residual nbits=4 | 185.0 | 64 | 8× | 0.7639 | — | ~21* | — | — |
| LateOn | residual nbits=2 | 108.7 | 32 | 16× | 0.7595 | 99.4% | 19.3 | 22.8 | ~1.1× |
| LateOn | **binary int8×1-bit** | **70.6** | **16** | **32×** | 0.7451 | 97.5% | **5.7** | 9.0 | **~3.7×** |
| LateOn-regularized | residual nbits=4 | 184.8 | 64 | 8× | 0.7615 | — | 21.1 | 25.2 | — |
| LateOn-regularized | residual nbits=2 | 108.5 | 32 | 16× | 0.7605 | 99.9% | 18.8 | 22.2 | 1.1× |
| LateOn-regularized | **binary int8×1-bit** | **70.4** | **16** | **32×** | 0.7513 | 98.7% | **5.6** | 8.9 | **3.8×** |

\* measured under CPU contention from a concurrent job; the regularized run
(byte-identical index layout) gives the clean 21.1 ms reference.

Summary: on ordinary 128-dim ColBERT checkpoints, `binary: true` retains
**97.5–98.7%** of residual NDCG@10 while making queries **3.8× faster**
(Stage-2 skips decompression entirely) and the index **2.6× smaller** than
nbits=4. nbits=2 compresses but barely changes latency — the speed win is
specific to scoring on the stored bits.

Known limitation, measured: low-dimension edge models are not binarizable
without help — `mxbai-edge-colbert-v0-17m` (dim=48) retains only 6% NDCG
(verified independently). The fused kernels cover those dims (48-dim scoring
runs at 0.94 µs/doc, not the 5.4 µs fallback), so the constraint is purely a
*quality* property of the checkpoint: for such models use residual nbits=2
(16×, ~100% retention) or a binarization-robust checkpoint.

## Surface exposure (colgrep + API, added during review)

Off by default everywhere; no behavior change unless opted in.

- **colgrep**: `colgrep settings --binary` / `--no-binary` — persisted in
  `~/.config/colgrep/config.json` and applied to every subsequent index build.
  Toggling re-embeds existing indexes on their next update. Binary indexes
  route added/changed files through an atomic full re-embed (they cannot take
  incremental appends); deletions stay incremental. Fresh binary builds run as
  a single batch so every write stays on the atomic initial-create path. The
  settings command warns that low-dim models (like the default 48-dim
  LateOn-Code-edge) may lose ranking quality.
- **next-plaid-api**: `POST /indices` accepts `"binary": true` in the config
  (stored, echoed, threaded into `IndexConfig`); matching Python SDK field.
- **next-plaid core**: `MmapIndex::update` now serves binary indexes through
  the start-from-scratch rebuild — preserving the storage scheme instead of
  silently reverting to residual — and rejects the buffer/centroid-expansion
  paths before any destructive step. So API document adds work while the index
  holds at most `start_from_scratch` documents; beyond that they fail with a
  clear error.
- Documented in both READMEs; regression tests cover update-via-rebuild
  keeping the flag, above-threshold rejection without mutation, and the
  settings round-trip.

## Testing

- 185/185 tests pass on native aarch64 and x86_64; integration tests cover
  storage width, retrieval, ±1 reconstruction, and update rejection.
- Fused kernels are pinned bit-exactly to a scalar integer reference across
  9 dims (8..256, whole/partial/odd-byte groups) × all 4-block tail shapes;
  `clippy -D warnings` and `fmt --check` clean on both architectures.


joe32140:
**Cross-platform kernel matrix — 3 ISAs × 4 OSes, all green**

To back the microbenchmark table with more than one machine, I ran a CI matrix on my fork ([run 29058497607](https://github.com/joe32140/next-plaid/actions/runs/29058497607), workflow on [`ci-kernel-matrix`](https://github.com/joe32140/next-plaid/tree/ci-kernel-matrix)). Each job runs the **full test suite in release mode** — the SIMD kernels are pinned bit-exact to the scalar reference, so this doubles as a per-architecture correctness gate — then `examples/kernel_bench`.

Per-doc Stage-2 latency (dim=128, 32 query × 80 doc tokens; speedup vs that host's per-pair fallback):

| runner | CPU | kernel | fused µs/doc | speedup |
|---|---|---|---:|---:|
| ubuntu-24.04 | Xeon Platinum 8370C (Ice Lake) | **AVX-512 VNNI** | 3.87 | 4.9× |
| — same host | — | AVX2 SAD | 6.34 | 3.0× |
| ubuntu-24.04 | EPYC 9V74 (AVX2 exposed) | AVX2 SAD | 4.86 | 4.0× |
| windows-latest | EPYC 7763 | AVX2 SAD | 4.64 | 5.0× |
| ubuntu-24.04-arm | Neoverse N2 | NEON SDOT | 4.84 | 7.7× |
| macos-latest | Apple M1 (virtual) | NEON SDOT | 2.47 | 8.0× |
| Apple M4 (local, native) | Apple M4 | NEON SDOT | 1.9 | 7.2× |

Observations:

- The dispatched default (`maxsim_binary_i8`) matched the best available kernel on every host.
- AVX2 masked-SAD lands within ~25% of VNNI in absolute µs — AVX-512-less x86 loses little.
- decode→f32-GEMM is *slower* than even the per-pair bit path on every x86 host tested (0.4–0.8×) — scoring the stored bits beats decompression even in the fallback.
- The EPYC 9V74 (a Genoa part) exposes only AVX2 inside the runner VM, so runtime feature dispatch — rather than compile-time assumptions — is doing real work even on AVX-512-era hardware.
- Windows/MSVC: full suite passes (`cargo test -p next-plaid --release`, default features).

Two changes pushed/edited alongside this comment:

- f0d09d4 adds `ubuntu-24.04-arm` to the CI test matrix: the NEON kernels this PR introduces are `cfg(target_arch = "aarch64")`-gated, so an x86-only matrix never compiles or executes them (arm64 runners are free for public repos). Happy to drop this commit or split it into its own PR if you'd rather keep this one focused.
- PR body: added a scale caveat under the end-to-end table (SciFact is a small, complete BEIR task — these numbers demonstrate codec correctness/retention, not large-corpus behavior), and the fallback recommendation for low-dim models now points at residual `nbits=1` (same 32× doc footprint as binary, 98.4% of nbits=4 NDCG on that checkpoint) rather than `nbits=2`.

raphaelsty:
Thank you for this MR, I'll review it asap :)

joe32140:
**Cross-platform validation of the dim generalization** (commit 21c1bf9)

The fused MaxSim kernels originally only kicked in at `dim == 128`; they now cover every byte-aligned dim up to 256 (`dim % 8 == 0`), with ragged dims and dims > 256 still taking the per-pair fallback. To validate, we ran the full test suite (kernels pinned bit-exact to the scalar reference) plus the `kernel_bench` sweep on five GitHub runners. All jobs green; fused numbers below (µs/doc, speedup vs the per-pair fallback on the same machine):

| dim | linux x86 (AVX2 SAD) | linux arm64 / Neoverse (NEON SDOT) | macos arm64 (NEON SDOT) | windows x86 (AVX2 SAD) |
|---:|---:|---:|---:|---:|
| 48 | 3.20 (18.3×*) | 2.27 (7.1×) | 1.30 (7.2×) | 5.89 (10.3×*) |
| 64 | 3.14 (4.5×) | 2.73 (7.8×) | 1.55 (7.3×) | 5.88 (2.8×) |
| 96 | 3.77 (4.8×) | 3.70 (8.2×) | 2.24 (7.1×) | 6.46 (3.2×) |
| 128 | 4.45 (5.2×) | 4.69 (8.6×) | 2.91 (7.0×) | 7.29 (3.4×) |
| 256 | 7.53 (5.2×) | 10.94 (7.1×) | 7.52 (5.6×) | 10.19 (4.2×) |

\* dim=48 packs to 6 bytes, which isn't u64-aligned — the per-pair fallback degrades to slow tail handling there (~60 µs/doc on x86), so the large ratios mostly reflect how much that case needed rescuing. dim=48 matters in practice: it's what edge ColBERT checkpoints like `mxbai-edge-colbert-v0-17m` ship at.

Notes:
- The fused kernel beats the fallback at **every dim on every ISA**, and the dispatched entry point (`maxsim_binary_i8`) matches the forced kernel within noise — the `dim % 8 == 0 && dim <= 256` gate is safe as the unconditional default.
- GitHub runner hardware is heterogeneous: this run's x86 draws lacked AVX-512 VNNI and exercised the AVX2 masked-SAD tier end-to-end; an earlier run drew a VNNI machine (3.23 µs at dim=128, 5.2×). Both x86 tiers are now hardware-validated.
- Apple M4 (native, not a runner) for reference: 0.94 / 1.16 / 1.54 / 1.98 µs at dims 48/64/96/128.

Reproduce: `cargo run --release -p next-plaid --example kernel_bench -- <dim>`.

raphaelsty:
Amazing PR!!! I'm on holiday for a few day, but I'll review and merge asap, thank you for this work and the benchmarks ☺️

joe32140:
No worries. Thanks!

joe32140:
Thanks for making commits! Please let me know if I can help with anything :)

raphaelsty:
@joe32140 Sorry it took a while, I think it's ready to merge, I took a bit of time to also expose the binary storage in colgrep settings and in next-plaid-api and add a dedicated section on the readme for this

PR #157, opened 2026-07-21
feat(colgrep): add first-class Dart support

amirhdrz:
## What

Adds Dart to ColGREP.
It uses https://github.com/nielsenko/tree-sitter-dart as the tree-sitter parser, which is not an official Tree-sitter parser.

Dart extraction covers:

- Top-level functions and methods
- Constructors, getters, setters, and operators
- Classes, mixins, extensions, extension types, and enums
- Modern and legacy typedefs
- Parameters, return types, `///` documentation, calls, variables, imports, and inheritance

The supported-language documentation and Cargo lockfile are updated as well.

## Implementation notes

- Uses `tree-sitter-dart 0.2.0` through the existing `LANGUAGE.into()` integration.
- Normalizes Dart's separate signature and function-body nodes so extracted units contain the complete implementation.
- Handles Dart's current `call_expression`, `member_expression`, and `class_declaration` grammar shapes.
- Prevents nested method signatures from producing duplicate code units.
- Preserves class and method relationships, including inheritance and qualified names.

## Testing

- `cargo check -p colgrep`
- `cargo test -p colgrep dart` 3 passed
- `cargo test -p colgrep` 722 passed, 1 pre-existing timing test ignored
- Doctests pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -616,6 +616,7 @@ dependencies = [
  "tree-sitter-cmake",
  "tree-sitter-cpp",
  "tree-sitter-css",
+ "tree-sitter-dart",
  "tree-sitter-elixir",
  "tree-sitter-go",
  "tree-sitter-graphql",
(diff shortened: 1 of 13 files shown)


raphaelsty:
Hi @amirhdrz, thank you for this MR, I was wondering why you did not choose this [specific crate](https://crates.io/crates/tree-sitter-dart-orchard) to handle dart code processing but I'm fine merging the MR like this. 

Thank you :)

amirhdrz:
Hi, thanks for approving this pull request.

The honest answer is that I had not considered tree-sitter-dart-orchard but it looks like a reasonable choice at first glance. My only criteria was that the latest Dart syntax is supported.

If there is interest I can help switch over.

PR #158, opened 2026-07-21
feat(colgrep): add Kimi Code agent integration

IgorCarron:
Adds --install-kimi / --uninstall-kimi for Kimi Code CLI, following the existing OpenCode/Codex/Hermes pattern.
 
Kimi Code discovers agent skills from $KIMI_CODE_HOME/skills/<name>/SKILL.md (default ~/.kimi-code/skills/). The installer writes the shared colgrep skill body there with the YAML frontmatter Kimi requires (name, description,  whenToUse), so new sessions automatically use colgrep as the primary search tool. 

colgrep --uninstall sweeps the Kimi skill too.
   
Verified: cargo check, cargo fmt --check, and an install/uninstall round-trip of the built binary (file written with correct frontmatter, removed cleanly).
   
Happy to rework this toward the generic --install-agent shape from #96 if preferred.


raphaelsty:
Thank you @IgorCarron for this MR, this is perfect 🐐🤌

PR #160, opened 2026-07-23
Fall back to CPU when automatic accelerator setup fails

MaxHalford:
## What changed

- add an opt-in CPU retry when accelerator-backed ONNX session creation fails
- enable the retry for colgrep's automatic acceleration mode
- keep `--force-gpu` strict
- record CPU as the effective provider after fallback

## Why

A cold CoreML model compilation can fail in restricted macOS environments even when the persistent cache directory is writable. In automatic mode, that currently prevents semantic search although CPU inference remains available.

## Validation

- `cargo test -p next-plaid-onnx --lib` (33 passed)
- `cargo test -p colgrep --features coreml --lib` (673 passed, 1 ignored)
- `cargo clippy -p colgrep --features coreml --all-targets -- -D warnings`
- cold-cache sandbox smoke test: automatic mode returned results through CPU fallback
- matching `--force-gpu` smoke test retained the CoreML compilation error

The implementation and this description were prepared with OpenAI Codex. The commit includes a `Co-authored-by` trailer.

### DIFF
--- a/colgrep/src/index/mod.rs
+++ b/colgrep/src/index/mod.rs
@@ -1340,6 +1340,7 @@ impl IndexBuilder {
                     .with_batch_size(batch)
                     .with_dynamic_batch(self.dynamic_batch)
                     .with_execution_provider(execution_provider)
+                    .with_cpu_fallback(env_acceleration_mode_lossy() == AccelerationMode::Auto)
                     .build()
             })
             .context("Failed to load ColBERT model")?;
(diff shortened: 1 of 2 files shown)


MaxHalford:
Disclaimer: This is completely vibe coded PR, following an issue we noticed with @raphaelsty, while installing colgrep in person on my laptop.

MaxHalford:
Superseded: the CPU fallback treated the symptom. Further controls isolated the issue to CoreML's cold-compilation working path, so I replaced this with a path-level fix that preserves CoreML execution.

PR #161, opened 2026-07-23
Use the CoreML cache as its compilation working directory

MaxHalford:
## What changed

- resolve the effective CoreML cache directory at colgrep startup
- create it before model loading
- use it for both `ModelCacheDirectory` and the process `TMPDIR`
- perform this setup before the global Rayon worker pool starts

## Root cause

CoreML cold compilation uses `TMPDIR` for intermediate work even when ONNX Runtime has a persistent `ModelCacheDirectory`.

In a macOS Seatbelt workspace, ordinary file writability is not sufficient for CoreML's compilation service:

- default `/var/folders/.../T`: process writes succeeded, CoreML compilation failed
- canonical `/private/var/folders/.../T`: CoreML compilation failed
- `/private/tmp`: CoreML compilation failed
- a directory covered by the workspace sandbox extension: CoreML compilation succeeded
- colgrep's writable CoreML cache directory: CoreML compilation succeeded

After an external cold compile populated the model cache, sandboxed loads also succeeded. This isolated the failure to CoreML's intermediate compilation path rather than model download, cache lookup, ONNX Runtime loading, repository indexing, or inference.

## Why this fix

The configured CoreML cache is already intended to be a stable writable location for compiled model artifacts. Reusing it as CoreML's working directory keeps cold compilation on an accessible path and preserves CoreML execution. There is no CPU fallback.

## Validation

- `cargo test -p colgrep --features coreml --lib` (673 passed, 1 ignored)
- `cargo clippy -p colgrep --features coreml --all-targets -- -D warnings`
- cold-cache `--force-gpu` smoke test in the failing Seatbelt workspace succeeded on CoreML without an external `TMPDIR` override

The implementation and this description were prepared with OpenAI Codex. The commit includes a `Co-authored-by` trailer.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2562,6 +2562,7 @@ name = "next-plaid-onnx"
 version = "1.6.4"
 dependencies = [
  "anyhow",
+ "fs2",
  "glob",
  "ndarray 0.16.1",
  "numpy",
--- a/colgrep/src/main.rs
+++ b/colgrep/src/main.rs
@@ -10,11 +10,13 @@ use anyhow::Result;
 use clap::{CommandFactory, Parser};
 use rayon::ThreadPoolBuilder;
 
+#[cfg(feature = "coreml")]
+use colgrep::Config;
 use colgrep::{
     acceleration::{apply_acceleration_mode, env_acceleration_mode, AccelerationMode},
     install_claude_code, install_codex, install_hermes, install_kimi, install_opencode,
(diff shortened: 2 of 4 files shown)


MaxHalford:
Disclaimer: This is completely vibe coded PR, following an issue we noticed with @raphaelsty, while installing colgrep in person on my laptop.

raphaelsty:
WIP :)

PR #162, opened 2026-07-23
Speed up cold indexing by 5–11x with adaptive context

cktang88:
## Impact

Cold indexing becomes **5.3x faster on React** and **11.3x faster on a large private repository** on an Apple M4 Pro with 48 GB RAM, while peak memory falls by 67–73%.

This revision replaces the earlier fixed 64-token policy with adaptive, structure-aware context budgets. That recovers substantially more retrieval quality, but the benchmark results below also show that the speedup is not quality-neutral on every workload.

## What changes for users

- Fresh indexes are encoded and written in one pass instead of repeatedly rebuilding index-wide IVF data at each checkpoint. Resumed partial indexes retain checkpointing.
- Encoding uses a bounded worker pool with at most one worker per ONNX session.
- The default code model selects a 64, 128, 256, or 512-token budget from the parsed code unit's type, span, control flow, and complexity. Short/simple units remain cheap; larger functions retain more context.
- JavaScript files marked `@flow` are parsed with the existing TSX grammar so typed functions remain coherent units.
- Function-local declarations are not separately embedded for JavaScript-family languages when their enclosing function already contains them.
- `COLGREP_INDEX_DOCUMENT_LENGTH=0` restores the model-native context. A positive value forces a fixed limit. Custom models continue to use their native limit by default.
- Existing indexes rebuild once because the embedding policy and index format changed.

## Cold-index benchmarks

Release binaries used the same base commit, cached model, CPU INT8 inference, CoreML + Accelerate support, 14 ONNX sessions, and isolated cold index directories.

### React

Pinned at [`react/react@28cd4bb`](https://github.com/react/react/commit/28cd4bb08f1b66808bede284fca978cc9b065154), indexing the same 6,959 files:

| Metric | `main` | This PR | Change |
| --- | ---: | ---: | ---: |
| Wall time | 196.67 s | 37.15 s | **5.29x faster** |
| CPU time | 1,380.17 s | 362.50 s | **3.81x less** |
| Peak RSS | 6.53 GB | 1.79 GB | **73% lower** |
| Parsed/indexed units | 90,602 | 52,903 | Coherent parsing removes duplicate/fragmented units |

### Private large repository

Both runs indexed the same 10,406 files:

| Metric | `main` | This PR | Change |
| --- | ---: | ---: | ---: |
| Wall time | 621.07 s | 54.75 s | **11.34x faster** |
| CPU time | 4,575.02 s | 509.88 s | **8.97x less** |
| Peak RSS | 6.49 GB | 2.18 GB | **66% lower** |
| Parsed/indexed units | 119,490 | 72,255 | Coherent parsing removes duplicate/fragmented units |

Only aggregate measurements are reported; no proprietary source, paths, fixtures, query text, or repository-specific behavior is included.

## Retrieval evaluation

The evaluation uses 100 recent non-merge commit subjects as queries and treats files changed by the corresponding commit as relevant. Searches use `--files-only -k 10`. This is a reproducible proxy for code-search usefulness, not a human relevance judgment.

### React

| Search mode | Branch | Recall@1 | Recall@5 | Recall@10 | MRR |
| --- | --- | ---: | ---: | ---: | ---: |
| Default/hybrid | `main` | 0.41 | 0.72 | 0.80 | 0.536 |
| Default/hybrid | This PR | **0.50** | **0.74** | **0.83** | **0.602** |
| Semantic only | `main` | 0.42 | **0.67** | **0.75** | 0.536 |
| Semantic only | This PR | **0.47** | 0.63 | 0.72 | **0.542** |

Default/hybrid search improves at every cutoff. Semantic-only search improves Recall@1 and MRR but loses 4 points at Recall@5 and 3 points at Recall@10.

### Private large repository

| Search mode | Branch | Recall@1 | Recall@5 | Recall@10 | MRR |
| --- | --- | ---: | ---: | ---: | ---: |
| Default/hybrid | `main` | **0.52** | **0.77** | **0.84** | **0.637** |
| Default/hybrid | This PR | 0.49 | 0.72 | 0.79 | 0.593 |
| Semantic only | `main` | **0.49** | **0.73** | **0.79** | **0.593** |
| Semantic only | This PR | 0.36 | 0.59 | 0.67 | 0.462 |

The private workload shows a modest default/hybrid regression and a material semantic-only regression. The native-context environment override remains available, but reviewers should treat this PR as an explicit performance/quality tradeoff rather than a claim of universal quality parity.

## Why cold builds were slow

Three costs compounded:

- Each checkpoint appended to a growing index and rebuilt index-wide IVF data, making a fresh build effectively quadratic.
- Each encoding chunk spawned one OS thread per batch even though only a bounded number of ONNX sessions could run.
- The default model accepted up to 2,048 document tokens, so a small long-document tail dominated transformer inference.

GPU/SIMD changes were not the primary lever on this workload: eliminating repeated index construction, thread oversubscription, fragmented units, and unnecessary transformer context produced the measured gains while retaining the existing cross-platform inference path.

## Manual QA

- [x] Compare isolated cold indexes on pinned React.
- [x] Compare isolated cold indexes on the private large repository.
- [x] Run the 100-query retrieval proxy in default/hybrid and semantic-only modes on both repositories.
- [x] Confirm native and fixed-length environment overrides remain available.
- [x] Confirm no proprietary source, paths, fixtures, or queries are present in the branch.
- [ ] Decide whether the private-repository semantic-only tradeoff is acceptable before merging.

### DIFF
--- a/colgrep/src/index/mod.rs
+++ b/colgrep/src/index/mod.rs
@@ -25,7 +25,9 @@ use serde::{Deserialize, Serialize};
 use crate::acceleration::apply_acceleration_mode;
 use crate::acceleration::{env_acceleration_mode_lossy, AccelerationMode};
 use crate::embed::build_embedding_text;
-use crate::parser::{build_call_graph, detect_language, extract_units, CodeUnit, Language};
+use crate::parser::{
+    build_call_graph, detect_language, extract_units, CodeUnit, Language, UnitType,
+};
 use crate::signal::{is_interrupted, is_interrupted_outside_critical, CriticalSectionGuard};
 
 use paths::{
@@ -56,11 +58,16 @@ const BUILDING_MARKER: &str = ".building";
 /// more often (more resumable) at the cost of more index-append overhead.
 const BUILD_CHECKPOINT_UNITS: usize = 4096;
 
-/// Keep the default code model's indexing context small. Embedding text puts symbol identity,
-/// signature, documentation, and path first, so these survive truncation; hybrid FTS still
-/// searches the full stored source. Set `COLGREP_INDEX_DOCUMENT_LENGTH=0` to use the model's
-/// native limit, or set a positive token count to choose a different quality/speed tradeoff.
-const DEFAULT_INDEX_DOCUMENT_LENGTH: usize = 64;
+/// Upper bound for adaptive indexing with the default code model. Most code units use much
+/// less based on their AST-derived shape; complex functions can use the full budget.
+const ADAPTIVE_INDEX_DOCUMENT_LENGTH: usize = 512;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum IndexDocumentLength {
+    Native,
+    Fixed(usize),
+    Adaptive,
+}
 
 fn build_checkpoint_units(binary: bool, index_exists: bool) -> usize {
     if binary || !index_exists {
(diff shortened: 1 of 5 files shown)


cktang88:
@raphaelsty any thoughts?

raphaelsty:
Hi @cktang88 sorry to answer a bit late, I'll review your MR asap 😊

raphaelsty:
Hi @cktang88 , thanks for the detailed benchmarks, and for providing honest benchmarks as well, it's nice to see the tradeoff

Could we split this MR in two? 

1st MR: The one-pass build and the bounded worker pool are easier to merge 

2nd MR: The adaptive context part worries me more: Recall@1 going from 0.49 to 0.36 in semantic-only mode
 on your private repo is a big hit for a default change. I'd rather try more generous thresholds I think so.  I totally get that it's much more faster so it could be something we could configure, I need to think about this tradeoff. Might be painful but is there any lever to increase recall on your private repository ?

PR #166, opened 2026-07-30
feat(colgrep): add persistent stdio search server

Acters:
## Summary

Add an opt-in persistent local search mode:

```text
colgrep serve --stdio --color=never --model MODEL -- ROOT
```

The server loads one existing index and reuses the same `SearchEngine`, tokenizer, ONNX session, filtering database, and PLAID mmap across requests. It uses stdin/stdout only: there is no network listener or background daemon.

## Protocol

The server speaks bounded version-1 NDJSON and supports:

- `health`
- `search`
- `shutdown`

Requests use scalar string or number IDs. Search accepts `query`, `top_k`, `semantic_only`, `code_only`, `include`, `exclude`, and a contained relative `restrict_to_dir`. Successful searches return the same serialized results and deterministic ordering as one-shot `colgrep search --json`.

Protocol safeguards include:

- protocol-only stdout; diagnostics remain on stderr
- 64 KiB request, 8 MiB response, 4096-byte query, and 100-result limits
- at most 16 glob patterns of 256 bytes each and 256 brace-expanded forms total
- malformed glob rejection and component-bounded directory filtering
- oversized input draining so the next NDJSON request remains synchronized
- explicit error responses instead of successful empty results when filtering or search fails

## Index consistency

The server uses the existing exclusive index lock for its full lifetime. `colgrep init` or another index update therefore requires shutting down the server first. Startup refuses missing, dirty, or in-progress indexes and never creates, repairs, or updates an index.

This keeps the consistency model small: the implementation reuses the normal search path and does not add a separate read-only loader, generation tracking, or index-lifecycle layer.

## Validation

- `cargo test --workspace --locked`
- `cargo clippy --workspace --all-targets --locked -- -D warnings`
- ordered one-shot/server parity for natural, exact, semantic-only, code-only, include, exclude, and restricted searches
- live checks for request bounds and resynchronization, glob semantics, directory-component containment, malformed configuration, and exclusive-lock release on shutdown

On the release build and current test index, ten warm persistent searches measured 70.1 ms median versus 430.2 ms median for fresh one-shot processes (about 6.1x faster). Server startup through `health` measured 460 ms, and protocol stderr remained empty.


Acters:
I was eventually going to create this PR, but the LLM published it without considering on asking me to approve it.

The damage was already done, I will not close this one because it can be useful to keep a stdio server running for those who wish to make multiple search queries in a single session with a client (such as an agentic harness).

raphaelsty:
Hi @Acters thank your for your MR, it's fine to publish it :)

I'll need a bit of time to review it but it is relevant, what matters to me is to make this feature in the smallest amount of linecode as possible as well

Acters:
I agree, I got busy need to figure out a bit more on it before it can be properly reviewed

Acters:
I had been working on integrating it into the harness that I am using for about a week now, and well the AI decided to make a PR pre-emptively. It needed a lot of changes that I had been making over the span of the week; This latest commit would have been what I would have eventually put here.

Compared to the initial semi-broken commit, this PR is now along with mariad of smaller fixes and architectural changes.

It is also:
- Reusing the existing searchEngine and one-shot search pipeline.
- Using one exclusive index lock for the server lifetime.
- Requiring shutdown before indexing or updating.
- Removing the separate read-only loader, generation tracking, and lifecycle machinery.
- Keeping the protocol client-agnostic and locally scoped.

It is also simpler, I could try to harden it but that would make this PR larger than it is now with the simpler agnostic system it is employing. For right now, I halved the number of lines. 
Should be good for a review now.

It is very helpful and accurate, thank you to the lightonai team.

Acters:
## Rebase and conflict resolution

Rebased the current minimal stdio implementation onto upstream main at `aeaa64c`, including the recently merged subdirectory indexing changes.

Conflict resolution preserved both sides' intended behavior:

- upstream excluded-directory detection and `colgrep init <path>` guidance now flows through the shared CLI/server search implementation;
- server requests remain protocol-only and do not emit CLI guidance;
- directory filtering retains the component-bounded, parameterized lexicographic range, which also avoids SQL `LIKE` wildcard behavior.

Validation completed successfully:

- `cargo test --workspace --locked`
- `cargo clippy --workspace --all-targets --locked -- -D warnings`
- formatting and diff checks
- ordered one-shot/server parity across natural, exact, semantic-only, code-only, include, exclude, and restricted searches
- live oversized-request recovery, directory-component scoping, exclusive-lock lifecycle, and excluded-directory guidance
- focused post-rebase conflict audit: no findings

**Commit:** `b9830724ee1b03f655e1a9b884bcd3c4970685e6`

---
*Progress tracked by mach6*

PR #169, opened 2026-08-02
feat: asymmetric residual scoring — int8 query × fused LUT MaxSim, 5–8× faster rescoring, identical NDCG@10

joe32140:
## What

Optional asymmetric scoring for **residual** indexes
(`SearchParameters::residual_asym`, default off): Stage-2 scores the
*stored* codes directly — int8 query × a fused byte→int8-weights table,
plus the centroid term Stage-1 already computed — instead of
decompressing every candidate token to f32 and running a BLAS MaxSim.
Compute-only: same index, same storage, so the two modes can be A/B'd
per search. The residual-codec counterpart of #155's int8 × binary
scoring. #170 is the stage-1 half of the same campaign; the two are
independent and can land in either order.

It is off by default and reachable from every shipped surface — Rust, the
HTTP API, the Python SDK and its CLI, and colgrep (see *Enabling it*
below). Every addition is additive and defaults to today's behaviour.

## How

Scoring splits the dot product exactly:

```
q · token = q · centroid[cid]                    (from the IVF probe matrix)
          + Σ_d q_d · bucket_weights[code_d]     (int8 × int8, integer MACs)
```

then applies the float path's own per-token renormalize via a cached
`1/‖centroid + residual‖` (computed once per index; skipping it measures
up to −0.17 NDCG@10 at nbits=1, which is why it exists).

For byte-aligned dims ≤ 256, fused doc-token-outer kernels expand each
token's packed bytes once in registers — the 256-entry byte→weights
table provably factors into per-key-position 16-entry nibble tables
(verified over all 256 byte values at build, scalar fallback if the
packing ever changes), the shape NEON `tbl` / SSE `pshufb` consume —
amortized over all query rows:

- **NEON** `tbl` expand + SDOT, epilogue folds 4 query rows per
  `vmaxq_f32` with a `vpaddq` transpose-reduce
- **AVX2** `pshufb` expand + `maddubs`/`madd`, 8-wide fold
- **AVX-512 VNNI** `vpdpbusd` (sign carried via `movepi8_mask` +
  `mask_sub_epi8`; exact because both operands clamp to ±127), 16-wide
  fold

The epilogue reads the centroid term from a **centroid-major** `[K, nq]`
matrix — Stage-1's `[nq, K]` matrix transposed once per query with a
cache-blocked pass — so one doc token touches one contiguous strip
instead of gathering `nq` floats a row apart. (A controlled per-component
ablation showed this layout, not the SIMD epilogue, is the dominant win
on every CPU tested.)

All paths compute the identical integer accumulator and the identical
float epilogue expression: the parity suite asserts **bit-equality** with
the scalar reference across nq × nbits × dim — each arch kernel called
directly, plus the dispatcher — and a semantics test pins normalized
scoring to the `decompress` reference. The batched-centroid path
(num_centroids > centroid_batch_size) packs its sparse centroid scores
into a compact centroid-major matrix with a per-doc code remap, so asym
scoring survives large-K indexes unchanged (regression-tested against the
dense path).

## Enabling it

Compute-only, so the same index serves both modes and you can A/B per request
without rebuilding anything.

**Rust**

```rust
let params = SearchParameters { residual_asym: true, ..Default::default() };
index.search(&query, &params, None)?;
```

**HTTP API** — under `params` in the search body:

```json
{"queries": [[[0.1, 0.2]]], "params": {"top_k": 10, "residual_asym": true}}
```

**Python SDK**

```python
from next_plaid_client.models import SearchParams

client.search("my_index", queries, params=SearchParams(top_k=10, residual_asym=True))
```

**Python CLI**

```bash
next-plaid search my_index --residual-asym "your query"
```

**colgrep** — a search-time env var alongside the three existing knobs:

```bash
COLGREP_RESIDUAL_ASYM=1 colgrep "your query"
```

The SDK field is `Optional[bool]` defaulting to `None`, and is omitted from the
request payload when unset rather than sent as `false` — so callers who express
no opinion keep tracking the server-side default if it ever changes.

> **If you are on Apple Silicon, check your toolchain first.** A rustup default
> host of `x86_64-apple-darwin` — the usual state on a Mac migrated from Intel —
> or an x86_64 container makes the binary run under Rosetta. Rosetta does not
> expose AVX2, and NEON cannot exist on an x86 target, so the rescore falls all
> the way to the scalar kernel: this PR then measures **~1.2× instead of ~5×**,
> with no error and correct results. Nothing about the flag is wrong in that
> case, and no build flag fixes it.
>
> ```bash
> rustup show                       # "Default host" must be aarch64-apple-darwin
> file target/release/<binary>      # must say arm64, not x86_64
> rustup override set stable-aarch64-apple-darwin   # per-directory fix
> ```
>
> This is the single most likely reason to see a disappointing number from this
> PR. The measured cost of getting it wrong is in the next section.

## Verifying the fast path actually ran

**Read this before benchmarking.** Kernel selection is a *runtime* check
(`is_x86_feature_detected!` / `is_aarch64_feature_detected!`), so no build
flags are needed on real x86 or ARM hardware — `-C target-cpu=native` is
not required and does not help. But when the check fails the code falls
back to a scalar kernel silently and still returns correct scores, so the
only symptom is that the speedup collapses:

| same machine, same index, same commit — only the build target differs | kernel | float | asym | ratio |
|---|---|---|---|---|
| native aarch64 | `neon-sdot` | 15.75 ms | 3.16 ms | **4.99×** |
| x86_64 under Rosetta | `scalar` | 27.76 ms | 22.75 ms | **1.22×** |

(SciFact, 5183 docs, nbits 4, 300 real queries.) **A reported ~1.2× is
this, not a property of the change.** The ways to land there: an x86_64
toolchain or x86_64 container on Apple Silicon, an ARM CPU without
`dotprod`, or a dim that is not a multiple of 8.

Two things in this PR make that self-diagnosing:

- The library warns **once per process** when `residual_asym` was asked
  for and the fused kernel will not run — either because dispatch is
  unavailable, or because the index is binary / above the dim ceiling.
  The success case stays silent unless you set
  `NEXT_PLAID_REPORT_KERNEL=1`, which reports the kernel name so a
  benchmark can record what produced its numbers.
- `examples/asym_rescore_check.rs` names the dispatched kernel, then
  times both Stage-2 arms. No dataset, no index build, about a second:

  ```
  cargo run --release -p next-plaid --example asym_rescore_check
  ```

  On an M4 that reports 3.0× with `neon-sdot`, and 0.72× — asym *slower*
  than float — when built x86_64 under Rosetta. Its synthetic corpus is
  RAM-resident, where the float arm costs ~44 ns/token against ~67 on a
  real mmap-backed index, so its ratio is a floor rather than a
  reproduction of the numbers below; its job is the kernel line. **Feel
  free to delete it once you have run it** — nothing else references it,
  so reverting that one commit removes it cleanly.

## Measured

- **Quality — NDCG@10 is unchanged**: |Δ| ≤ 0.002 vs the float path on
  identical codes, and identical to three decimals at deployed settings —
  3 ColBERT checkpoints × 3 BEIR corpora × nbits 4/2/1, incl. long-query
  ArguAna. The int8 error lands only on the residual correction; the
  dominant centroid term stays float. Spot-checked again on real GTE
  embeddings at deployed settings (seed-42 builds, float vs asym on
  identical indexes):

  | dataset | r=4 | r=2 | r=1 |
  |---|---|---|---|
  | NFCorpus | 0.3809 / 0.3811 | 0.3779 / 0.3779 | 0.3701 / 0.3705 |
  | SciFact | 0.7609 / 0.7607 | 0.7507 / 0.7507 | 0.7427 / 0.7427 |

  Max |Δ| across the six cells: 0.0004.
- **Latency** — the rescore stage (exact scoring of ~1024-doc
  shortlists, real corpus shapes), fused asym vs the float
  decompress+GEMM path on identical indexes. The CI rows come from one
  interleaved run of the same tree with `residual_asym` off vs on —
  same stage-1, same shortlists, so the ratio isolates this PR's
  scheme change. 9 cells per platform (3 datasets × nbits 4/2/1);
  geomean, with the per-cell range in parentheses:

  | platform | fused vs float rescore | decompress share of the float rescore |
  |---|---|---|
  | x86 AVX2 (GitHub CI) | **6.1×** (5.2–7.8×) | 74–83% |
  | Neoverse (GitHub CI) | **6.3×** (5.3–8.2×) | 67–76% |
  | macOS arm64 (GitHub CI) | **5.1×** (4.3–5.7×) | 71–80% |
  | Apple M4 (native, 0.5–7M-token corpus ladder) | **5.2×** (4.7–5.8×) | 48–72% |

  The right column is the mechanism: the fused kernels win by skipping
  decompression, not by out-multiplying GEMM — and it also bounds the
  win by whatever share decompression holds in a given deployment (the
  share falls as corpora grow: 72% → 48% across the M4 ladder).
- **End-to-end**: this PR alone moves whole-query latency 1.4–5.2×
  depending on corpus size — stage-1 dominates at scale, and that is
  #170's job. Combined with the stage-1 rework, an interleaved A/B
  against v1.6.5 over 18 cells (2 datasets × nbits 4/2/1 × 3 platforms)
  measures **7.03× end-to-end geomean** — 6.59× x86 AVX2, 7.08×
  Neoverse, 7.46× macOS arm64. That combined number belongs to the pair,
  not to this PR alone.
- **AVX-512 honesty note**: the VNNI kernel is written, feature-gated,
  and covered by the parity suite *on VNNI hardware*, but GitHub's
  standard runners don't have VNNI — correctness-validated, no perf
  claim.

Measurement details, per-component ablations, and the bench harness live
on the research branch `feat/asymmetric-lut-residual`.

## Should this be the default?

It is default-off here, deliberately: the asym path is quantized, so while
NDCG@10 matches, absolute *score values* shift slightly (28 of 300 SciFact
queries had a different top-1 score in the run above). Anyone with score
thresholds or snapshot tests would see that, and flipping a default that
changes returned scores felt like your call rather than something to bury
in a perf PR.

That said, the fallbacks are all safe — binary indexes, oversized dims, and
non-factoring codecs each return the float path untouched — so if you are
comfortable with the score-value caveat, defaulting it on is a one-line
change and I am happy to make it here or in a follow-up.

## Prior art

The mechanism — scoring quantized codes in registers instead of
decompressing them — is the asymmetric-distance-computation lineage:
FAISS FastScan / [André et al., VLDB 2016](http://www.vldb.org/pvldb/vol9/p288-andre.pdf)
established the SIMD-register nibble-LUT (`pshufb`) form for PQ, and
[WARP (SIGIR 2025)](https://arxiv.org/abs/2501.17788) applied "implicit
decompression" to late-interaction retrieval on CPU for XTR's PQ-style
codec. This PR is the counterpart for **PLAID's per-dimension scalar
residual codec**: rather than a precomputed-distance table (which fits
PQ subspace codebooks), it expands codes to int8 *weights* in registers
and runs an explicit int8 dot (SDOT / `maddubs` / `vpdpbusd`), which is
what lets it reuse Stage-1's centroid term exactly and stay bit-compatible
with the float path's renormalized semantics.


joe32140:
## Memory/latency follow-up

Commit b5661b7 removes the full-index inverse-norm cache added by this PR. Instead, exact scoring computes norms only for each shortlisted document into reusable worker-local scratch storage. The SIMD/LUT kernels and scoring math are unchanged.

This supersedes the PR description's "computed once per index" wording: norms are now computed on demand for shortlisted documents.

### Memory impact

- Removed persistent allocation: 4 bytes per indexed token.
- SciFact test index (1,191,618 tokens): 4.55 MiB modeled cache eliminated.
- PR #26-scale deployment (about 1.18B tokens): about 4.40 GiB persistent RSS avoided.
- The centroid-score transpose remains transient and bounded by K × query_tokens × 4 bytes: 8 MiB/query in this test (K=16,384, nq=128), or 32 MiB/query at K=65,536 and nq=128. At concurrency 16 that production shape models to 512 MiB.

### Linux ARM64 fresh-process result

Same SciFact r4 index, fixed 128-token queries, five timed batches after one warm-up, RAYON_NUM_THREADS=batch:

| concurrency | mainline RSS / latency | previous PR RSS / latency | bounded fix RSS / latency | fix vs mainline | fix vs previous PR |
|---:|---:|---:|---:|---:|---:|
| 1 | 93.52 MiB / 912.386 ms | 122.79 MiB / 701.447 ms | 98.25 MiB / 808.811 ms | 1.13× faster | 15.3% slower |
| 4 | 139.99 MiB / 235.014 ms | 171.95 MiB / 186.232 ms | 166.52 MiB / 209.961 ms | 1.12× faster | 12.7% slower |
| 16 | 261.73 MiB / 432.141 ms | 374.88 MiB / 379.506 ms | 369.72 MiB / 389.922 ms | 1.11× faster | 2.7% slower |

RSS includes mmap residency and runner noise, so the structural 4-byte/token removal is the scalable result; it is exact independently of the small-index RSS delta.

The cross-platform measurement run is green on Linux x86, Linux ARM64, and macOS ARM64: https://github.com/joe32140/next-plaid/actions/runs/31519800235. It benchmarked the same executable code at 6b2c558; b5661b7 differs only by a rustdoc markup correction.

Decision: keep the transpose for now because the bounded fix remains faster than mainline on ARM, while eliminating the unbounded memory/cold-start issue. The transpose should be evaluated separately against the target device's K, query length, and concurrency budget rather than removed from this focused change. residual_asym remains default-off.

Validation: full local next-plaid library/integration suite, clippy with warnings denied, rustdoc with warnings denied, and all PR CI/CLI E2E checks pass on final SHA b5661b7.

joe32140:
Implemented the bounded-memory fix for asymmetric residual scoring in `30f3916` and reran the combined PR #169 + PR #170 end-to-end benchmark against both mainline and the historical combined implementation.

### What changed

- Compute inverse residual norms once while indexing.
- Store them in an optional `f32` NPY sidecar and mmap the merged sidecar at search time.
- Borrow only the shortlisted token slice during LUT rescoring; there is no whole-index anonymous `Vec<f32>` allocation.
- Keep a correctness-compatible recomputation fallback for legacy indexes without the sidecar.
- Maintain the sidecar through index append/update/delete operations and validate its row count when loading.

### Corrected CI benchmark

Run: https://github.com/joe32140/next-plaid/actions/runs/31539529744

Matrix: NF-Corpus + SciFact, residual nbits 4/2/1, 3 repetitions in rotated order, Linux x64 + Linux ARM64 + macOS ARM64 (18 aggregate cells / 162 timed processes). The workflow explicitly passed `{"residual_asym":true}`, asserted sidecars existed, and enabled kernel reporting.

LUT dispatch was confirmed in the logs:

- Linux x64: 36 × AVX2
- Linux ARM64: 36 × NEON-SDOT
- macOS ARM64: 36 × NEON-SDOT

| Scope | Historical combined vs main | Sidecar combined vs main | Sidecar latency vs historical | Mean peak RSS delta vs historical |
|---|---:|---:|---:|---:|
| Overall | 6.162× | 6.057× | +1.7% | -3.38 MiB |
| Linux ARM64 | 6.633× | 6.485× | +2.3% | -3.73 MiB |
| Linux x64 | 7.153× | 7.029× | +1.8% | -3.86 MiB |
| macOS ARM64 | 4.931× | 4.875× | +1.1% | -2.55 MiB |

Across all cells, the RSS geometric-mean ratio is 0.952, i.e. about 4.8% lower than the historical combined implementation. The prior recompute-on-every-search version was about 1.8× vs main; the mmap sidecar restores that to about 6.1× while retaining the bounded/file-backed memory behavior.

### Validation

`cargo test -p next-plaid` passes: 149 unit tests plus every non-ignored integration suite. New coverage verifies sidecar values, sidecar/legacy score and ranking equivalence, update/delete consistency, row-count validation, mmap teardown, and preservation of the existing merged NPY layout.

### Deployment note

This is not a blocker for new/rebuilt indexes. Existing indexes remain correct and bounded-memory through the legacy fallback, but they will retain the slower recomputation path until rebuilt or backfilled with the inverse-norm sidecar. Edge deployments expecting the recovered ~6× speed should therefore include an index rebuild/backfill step.

joe32140:
Added legacy-index LUT prewarm support in `0506db5`.

`MmapIndex::prewarm_residual_lut_sidecar()` turns an existing residual index into the mmap-sidecar format without re-embedding or rebuilding the index:

```rust
let mut index = MmapIndex::load("/path/to/index")?;
index.prewarm_residual_lut_sidecar()?;
```

The method streams one existing codes/residuals chunk at a time, atomically writes only missing or length-mismatched inverse-norm chunks, performs the usual locked merge, validates the final row count, and immediately switches the live index to the mmap sidecar. It creates no full-index heap cache.

Operationally, call it during provisioning (before serving asymmetric residual queries). It is a no-op for binary indexes and indexes that already have the sidecar. It writes next to the index, so a future background/automatic mode can use an external cache for read-only deployments; that policy is intentionally not enabled here.

Validation:

- `cargo test -p next-plaid` — 149 unit tests and all non-ignored integration suites pass
- `cargo clippy --all-targets --features openblas -- -D warnings`

The legacy parity test now verifies: no-sidecar fallback results, prewarm sidecar creation/mmap, and post-prewarm LUT results all match.

joe32140:
Follow-up: hardened legacy prewarm memory in `def0cb1`.

The first prewarm implementation was bounded by an index-creation chunk. With the default 50,000-document chunk size, that could still mean millions of tokens and an unacceptable transient heap footprint on edge devices.

Prewarm now reads the existing merged code/residual mmaps and writes each legacy sidecar in fixed 64K-token blocks. At 128 dimensions and 4-bit residuals, the dominant payload touched per block is about 4 MiB, plus about 0.75 MiB of temporary code/inverse-norm buffers. Peak heap is therefore independent of the original chunk size and index size.

The one-time operation is still linear in token count and CPU/disk bound, so it should run during provisioning. It avoids the original implementation's whole-index `Vec<f32>` and avoids the previous prewarm implementation's full-chunk arrays.

Revalidated after the change:

- `cargo test -p next-plaid` — 149 unit tests plus all non-ignored integration suites pass
- `cargo clippy --all-targets --features openblas -- -D warnings`

joe32140:
Local release-mode prewarm throughput measurement for the inverse-norm kernel:

```
1,000,000 tokens, 128D, 4-bit residuals
184 ms = 5.42M tokens/s
```

CPU-only scaling from that measurement is about 1.11 s for 6M tokens and 1.85 s for 10M tokens.

This intentionally measures the real norm computation, not full sidecar I/O. Full prewarm also reads codes/residuals and writes then merges the sidecar. For a 6M-token, 4-bit index that moves roughly 500 MB total, so local end-to-end time will be somewhat higher; on edge storage/CPU it should still be treated as provisioning work, not a request-path operation.

The new 64K-token block implementation keeps the working payload bounded at about 4.75 MiB at this shape, independent of index size or the original 50K-document chunk size.

raphaelsty:
I made asymmetric residual scoring the default behaviour to keep things simpler after having carefully benchmarked it. Merging the MR, congrats for this work @joe32140

joe32140:
Thank you for testing it @raphaelsty. I am happy to see it as default!

PR #170, opened 2026-08-02
perf: rework the stage-1 shortlist pipeline — 4–12× faster stage 1 (5.6–7.1× e2e combined with #169)

joe32140:
Stage 1 and stage 2 are the two halves of a PLAID query. #169 rewrites the
stage-2 rescore; this PR rewrites stage 1. Together they take end-to-end search
**7.03x geomean** faster than v1.6.5 — 6.59x on x86 AVX2, 7.08x on Neoverse,
7.46x on macOS arm64, over the same 18 cells measured below.

**This PR stands alone.** It applies directly to v1.6.5, needs nothing from
#169, and is worth **1.30x geomean end-to-end on its own** (1.14–1.42x by platform) — the two can
land in either order.

## What

Stage 1 — everything a query pays before exact scoring — rebuilt phase by
phase, plus per-doc parallelism in the exact-scoring loop. Default-on: no new
parameters, no storage or API change. Every phase change ships with an
equivalence test against the path it replaces.

- **centroid GEMM**: column-block-parallel; the dim-long reduction stays inside
  one block, so the output is **bit-identical** to the single `dot` call
  (tested).
- **IVF probe**: running-threshold chunk scan replaces a K-length buffer fill +
  `select_nth` per query token — one sequential read per row, and chunks that
  can't beat the current k-th value are skipped on one predictable branch. Same
  top-k set by value (tested under ties).
- **candidate gather**: doc ids are dense, so a word bitmap dedups in
  O(postings) and scanning its set bits emits the same sorted list the old
  sort+dedup produced, without the sort.
- **approximate flood**: centroid scores are quantized to u8 in a fused
  transpose+quantize pass (PLAID's own centroid-interaction trick) and scored
  chunk-outer with a register-resident 16-lane max accumulator. The quantization
  is monotone per lane (u8 max = f32 max); the flood only ranks a 4096-deep
  prune cut. Gated on quality, not just kernel math: deployed nDCG@10 identical
  **to four decimals** vs exact flooding on real-embedding corpora (NFCorpus +
  SciFact, r4/r2/r1 and binary).
- **prune**: partial-select the surviving `n_full_scores` before sorting only
  that prefix; identical top list and order.
- **code reads**: a lazily-built u32 side-array of the token codes, so the flood
  reads 4 B/token instead of the mmap's 8. Built once on first search — the same
  lazy-derived-data pattern the index already uses.
- **exact scoring**: plain per-doc `par_iter` replaces fixed 128-doc chunking.
  The chunking's memory rationale no longer holds (in-flight decompressed docs
  are bounded by the thread count either way) and 8 chunks at n_decompress =
  1024 underfilled the pool — 6.1 effective threads measured on a 10-core M4;
  9.4 after.

## Measured

Interleaved same-box A/B against a v1.6.5 checkout with the same bench example
grafted in, identical prebuilt indexes, alternating main/ours per cell so
shared-runner drift hits both variants alike. 2 datasets x nbits 4/2/1 on each
of the three CI runner platforms (18 cells). Run on this branch as it stands,
with no part of #169 present:

| platform | stage 1 | end-to-end |
|---|---|---|
| x86 AVX2 (ubuntu-latest) | **4.78x** (4.72–4.85) | 1.14x |
| Neoverse (ubuntu-24.04-arm) | **8.39x** (6.45–10.75) | 1.35x |
| macOS arm64 | **7.31x** (7.00–8.04) | 1.42x |
| **all 18 cells** | **6.64x** (4.72–10.75) | **1.30x** (1.08–1.64) |

Geomeans. Absolutes on a shared runner are noise; the per-platform ratios are
the verdict. The same harness run against the stacked #169 + #170 tree
puts this PR's stage-1-only contribution at 1.13x / 1.33x / 1.50x on the three
platforms — reproducing the isolated numbers above (1.14x / 1.35x / 1.42x) from
a different tree, and giving the 7.03x combined figure quoted at the top. Server parts pay more for the scatter this removes than Apple's L2
does — which is why the native-M4 figure below is the conservative one.

Native Apple M4, SciFact 5,183 docs / 16,384 centroids / nbits 4, all 300 real
queries, timing the same span on both sides (cdot + probe + gather + flood +
prune):

| | stage 1 median |
|---|---|
| v1.6.5 | 6.649 ms |
| this branch | **1.693 ms** |
| | **3.93x** |

The biggest per-phase ratio (probe, 9.5x) is deliberately not the headline:
phase speedups compose by time share, and the flood + GEMM dominated. Per-phase
tables and the full exploration ledger (including the ideas measured and killed)
live on the research branch `feat/asymmetric-lut-residual`.

## Ranking

All 300 real SciFact queries return **identical top-10 ids, identical order, and
identical scores** vs v1.6.5 on a real-embedding index.

To be precise about what is and isn't guaranteed: five of the six phases are
exact-equivalent by construction and tested as such. The flood is not — it is
quantized, so its guarantee is nDCG parity (identical to four decimals), not
bit-identity. Where the quantization step is large relative to the gaps between
candidates, near-ties can reorder.

That shows up in the CI harness above, which drives *synthetic* indexes with
random unit-row queries: there roughly half the per-query digests differ between
the two builds (145 identical / 155 differing across 300 comparisons). Random
queries in 128-d are near-orthogonal to every centroid, so their scores cluster
inside a single quantization bucket and the ordering among them is arbitrary.
Real queries have score spread far above the quantization step, which is why the
300/300 result holds on real data. Worth knowing before reading the harness's
digests as a correctness signal — they are not one for this change.

Full suite: 205 tests green, `clippy -D warnings` and `cargo fmt --check` clean
on **each commit individually**.

## Relationship to #169

#169 is the stage-2 counterpart and the source of the combined figure above.
They are otherwise independent: this PR adds no scoring path and no public API
beyond a `codes_u32()` accessor, and #169 needs nothing from here.

One optimization was deliberately left out to keep this PR self-contained: when
#169's asymmetric arm is on, the flood's transpose pass can also emit the f32
centroid-major matrix stage 2 wants, so one traversal serves both stages. That
fusion only fires when `residual_asym` is set, so it is worth nothing here — it
belongs in whichever of the two PRs lands second.

### DIFF
--- a/next-plaid/src/index.rs
+++ b/next-plaid/src/index.rs
@@ -1062,6 +1062,12 @@ pub struct MmapIndex {
     pub mmap_codes: crate::mmap::MmapNpyArray1I64,
     /// Memory-mapped residuals array (public for search access)
     pub mmap_residuals: crate::mmap::MmapNpyArray2U8,
+    /// Lazily-built u32 copy of the token codes for the stage-1 candidate
+    /// flood: the mmap stores i64 (8 B/token), so a flood pass over millions
+    /// of candidate tokens reads 4x more code bytes than the values need
+    /// (num_centroids always fits u32). RAM cost is 4 B/token, paid on first
+    /// search only, and freed with the index.
+    codes_u32: std::sync::OnceLock<Vec<u32>>,
 }
 
 impl MmapIndex {
@@ -1183,24 +1189,55 @@ impl MmapIndex {
             doc_lengths,
             doc_offsets,
             mmap_codes,
+            codes_u32: std::sync::OnceLock::new(),
             mmap_residuals,
         })
     }
 
-    /// Get candidate documents from IVF for given centroid indices.
-    pub fn get_candidates(&self, centroid_indices: &[usize]) -> Vec<i64> {
-        let mut candidates: Vec<i64> = Vec::new();
+    /// Token codes as u32, built once on first use (one sequential pass over
+    /// the code mmap). Indexed by the same `doc_offsets` as `mmap_codes`.
+    pub fn codes_u32(&self) -> &[u32] {
+        self.codes_u32.get_or_init(|| {
+            let n = self.doc_offsets.last().copied().unwrap_or(0);
+            self.mmap_codes
+                .slice(0, n)
+                .iter()
+                .map(|&c| c as u32)
+                .collect()
+        })
+    }
 
+    /// Get candidate documents from IVF for given centroid indices.
+    ///
+    /// Doc ids are dense in `0..num_docs`, so a word bitmap dedups in
+    /// O(postings) and scanning its set bits emits the same sorted, deduped
+    /// list a `sort_unstable` + `dedup` of the concatenated postings would —
+    /// without the O(n log n) sort (measured 4–5x on this phase at 52k docs).
+    pub fn get_candidates(&self, centroid_indices: &[usize]) -> Vec<i64> {
+        let num_docs = self.doc_lengths.len();
+        let mut words = vec![0u64; num_docs.div_ceil(64)];
+        let mut count = 0usize;
         for &idx in centroid_indices {
             if idx < self.ivf_lengths.len() {
                 let start = self.ivf_offsets[idx] as usize;
                 let len = self.ivf_lengths[idx] as usize;
-                candidates.extend(self.ivf.slice(s![start..start + len]).iter());
+                for &d in self.ivf.slice(s![start..start + len]).iter() {
+                    let d = d as usize;
+                    let w = &mut words[d / 64];
+                    let bit = 1u64 << (d % 64);
+                    count += usize::from(*w & bit == 0);
+                    *w |= bit;
+                }
             }
         }
-
-        candidates.sort_unstable();
-        candidates.dedup();
+        let mut candidates: Vec<i64> = Vec::with_capacity(count);
+        for (wi, &word) in words.iter().enumerate() {
+            let mut w = word;
+            while w != 0 {
+                candidates.push((wi * 64 + w.trailing_zeros() as usize) as i64);
+                w &= w - 1;
+            }
+        }
         candidates
     }
 
--- a/next-plaid/src/index.rs
+++ b/next-plaid/src/index.rs
@@ -1199,11 +1199,7 @@ impl MmapIndex {
     pub fn codes_u32(&self) -> &[u32] {
         self.codes_u32.get_or_init(|| {
             let n = self.doc_offsets.last().copied().unwrap_or(0);
-            self.mmap_codes
-                .slice(0, n)
-                .iter()
-                .map(|&c| c as u32)
-                .collect()
+            (0..n).map(|i| self.mmap_codes.get(i) as u32).collect()
         })
     }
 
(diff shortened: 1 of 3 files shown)


joe32140:
### Memory and latency follow-up

I profiled current head [`45f1434`](https://github.com/lightonai/next-plaid/commit/45f1434dac366f9478e1d85ae73e06dfa39cb1cc) against mainline using 128-token queries, a 1,191,618-token / 16,384-centroid index, and fresh processes at concurrency 1, 4, and 16. Warm-up was excluded from latency; peak RSS still includes every retained allocation.

The new commit removes the lazy full-index `OnceLock<Vec<u32>>` copy and scores zero-copy from the existing code mmap. The u8 transpose remains enabled.

| Platform / concurrency | Mainline RSS | Current RSS | RSS delta | Latency speedup |
|---|---:|---:|---:|---:|
| Linux ARM ×1 | 93.39 MiB | 92.30 MiB | −1.09 MiB | 2.61× |
| Linux ARM ×4 | 140.54 MiB | 140.25 MiB | −0.29 MiB | 2.59× |
| Linux ARM ×16 | 261.20 MiB | 359.79 MiB | **+98.59 MiB** | 3.71× |
| Linux x86 ×1 | 93.57 MiB | 92.61 MiB | −0.96 MiB | 1.48× |
| Linux x86 ×4 | 139.57 MiB | 140.68 MiB | +1.11 MiB | 3.43× |
| Linux x86 ×16 | 261.40 MiB | 377.23 MiB | **+115.83 MiB** | 4.10× |

At the 1.18B-token scale, removing the cache eliminates a projected **4.40 GiB of persistent heap memory**. The remaining u8 transpose is approximately 8 MiB per in-flight query at K=65,536 and nq=128, or 128 MiB at concurrency 16 before allocator and working-set amplification.

My conclusion:

- This is not a merge blocker for edge deployments capped at concurrency 1–4; memory there is effectively mainline-equivalent while latency remains substantially faster.
- It is still a blocker to claiming unrestricted high-concurrency memory neutrality. Edge deployments should have an explicit concurrency or memory budget.
- I would keep the u8 transpose rather than discard the measured speedup, then address its concurrent transient memory separately.

The full three-platform benchmark is [here](https://github.com/joe32140/next-plaid/actions/runs/31515130568). All regular PR checks also pass on the current head.

raphaelsty:
Hi @joe32140 thank you for this benchmark, I like to see perfs and memory usage as well, this is very interesting. I'll merge this MR tomorrow :)

joe32140:
Follow-up for old-index compatibility, now pushed in `58d0165`.

PR #170 does **not** change the index format or break old indexes. It removes the whole-index `codes_u32` heap cache and reads the existing `merged_codes.npy` `i64` mapping instead.

I found that the project's historical merged-NPY header layout can leave the code payload unaligned for `i64`. The zero-copy cast correctly declines that layout, but the previous fallback allocated a `Vec<i64>` for every candidate document during stage 1.

This commit keeps the aligned zero-copy path and changes the unaligned fallback to decode codes directly from the mmap during the flood scorer. That preserves correctness and bounded memory without per-document allocations, so old indexes are immediately usable without a rebuild.

Added regression coverage that writes the historical unaligned header layout, verifies the zero-copy cast is rejected, and verifies the safe fallback reads the expected codes.

Validation:

- `cargo test -p next-plaid` — 147 unit tests plus all non-ignored integration suites passed
- `cargo clippy --all-targets --features openblas -- -D warnings`

PR #176, opened 2026-08-12
fix(onnx-export): support non-ModernBERT ColBERT models (LFM2, Jina-v2)

### LINKED ISSUE #93: pylate-onnx-export fails on several PyLate-compatible models (LFM2, Jina-v2) — multiple small bugs, proposed fixes
Trying to export `LiquidAI/LFM2-ColBERT-350M` (and `jinaai/jina-colbert-v2` as a fallback) to ONNX for use with ColGREP via `pylate-onnx-export` (package `colbert_export` v0.1.0, pulled from PyPI). 

The exporter only produces `tokenizer.json` + `config_sentence_transformers.json` — no `model.onnx` — and exits with a series of errors. Each error below was resolved by a small patch; the final blocker is architectural and specific to LFM2.

Environment: Python 3.12, `pylate-onnx-export==0.1.0`, `torch==2.x`, Linux.

## Bug 1 — `KeyError: 'token_type_ids'` on any non-ModernBERT architecture

```
Model architecture: Lfm2Model
Uses token_type_ids: True
Saved tokenizer to: ...
Saved config to: ...
Error: 'token_type_ids'
```

`detect_model_architecture` in `colbert_export/export.py` has a hardcoded allowlist:

```python
uses_token_type_ids = True
if "ModernBert" in model_class_name:
    uses_token_type_ids = False
```

Any non-ModernBERT backbone (LFM2, XLM-Roberta variants, Qwen, Llama-based ColBERTs…) defaults to `True`, then the exporter does `inputs["token_type_ids"]` against a tokenizer that didn't emit that key → KeyError.

**Proposed fix:** probe the tokenizer directly — authoritative for any architecture:

```python
tokenizer = pylate_model[0].tokenizer
probe = tokenizer("probe", return_tensors="pt")
uses_token_type_ids = "token_type_ids" in probe
```

## Bug 2 — new `torch.onnx` dynamo path fails on non-stock architectures

With bug 1 fixed, the export proceeds to `torch.onnx.export` and fails:

```
RuntimeError: 8*s72 (…) is not tracked with proxy for
  <torch.fx.experimental.proxy_tensor._ModuleStackTracer object at 0x...>
[torch.onnx] Obtain model graph for `ColBERTForONNX([...]` with
  `torch.export.export(..., strict=False)`... ❌
[torch.onnx] Obtain model graph for `ColBERTForONNX([...]` with
  `torch.export.export(..., strict=True)`... ❌
```

Recent PyTorch defaults the exporter to the dynamo path (`torch.export.export` → `onnxscript`), which doesn't handle some shape-dependent control flow in non-stock modeling code.

**Proposed fix:** pass `dynamo=False` to `torch.onnx.export` to fall back to the legacy TorchScript tracer, which is far more forgiving. Optionally make it configurable.

## Bug 3 — missing `onnxscript` dependency not declared

Even if you keep the dynamo path, `onnxscript` is needed and not declared in the package deps:

```
Error: No module named 'onnxscript'
```

**Proposed fix:** add `onnxscript` to install_requires (or at minimum document it in the README prerequisites).

## Bug 4 — no `trust_remote_code=True` for models that ship custom modeling code

When trying Jina-v2 as an alternative:

```
Error: jinaai/xlm-roberta-flash-implementation You can inspect the repository
  content at https://hf.co/jinaai/jina-colbert-v2.
Please pass the argument `trust_remote_code=True` to allow custom code to be run.
```

`pylate_models.ColBERT(...)` accepts `trust_remote_code`, but the exporter doesn't forward it. Blocks any HF model that ships custom modeling code (Jina, a bunch of community ColBERTs).

**Proposed fix:**

```python
pylate_model = pylate_models.ColBERT(
    model_name_or_path=model_name,
    device="cpu",
    do_query_expansion=False,
    trust_remote_code=True,
)
```

(Or expose it as a CLI flag, defaulting true — the user already has to opt in to running the exporter on this specific model.)

## Bug 5 — BF16 weights cause symbolic-registry misses during ONNX graph build

On Jina-v2 with all the above fixed, the full 24-layer graph traces, then fails at `linalg_vector_norm`/`clamp_min`/`aten::add` inside `F.normalize`:

```
Error: Argument passed to at() was not in the map.
```

(This is `std::unordered_map::at()` throwing `out_of_range` — the ONNX symbolic registry has no handler for the op/dtype combination.)

Root cause: Jina's `xlm-roberta-flash-implementation` loads in BFloat16; several ONNX symbolic handlers (at least through opset 20) have no BF16 coverage for these ops.

**Proposed fix:** force FP32 before export:

```python
model = ColBERTForONNX(pylate_model, uses_token_type_ids=arch_info["uses_token_type_ids"])
model = model.float()
model.eval()
```

After this patch, Jina-v2 still fails (see bug 6), but this unblocks any model that loads in BF16 by default.

## Bug 6 — architectural blocker on LFM2 and Jina's flash-XLM-Roberta (not actionable upstream)

With all above patches applied, both models still hit variants of `Argument passed to at() was not in the map.`:

- **LFM2 (`LiquidAI/LFM2-ColBERT-350M`):** fails inside Liquid-Foundation-Model blocks. Liquid's gated convolution and linear-recurrence ops have no ONNX equivalents at all — moving handlers to a higher opset doesn't help.
- **Jina-v2 (`jinaai/jina-colbert-v2`):** fails inside the XLM-Roberta-flash encoder at seemingly ordinary ops (`aten::linear`, `aten::add`, `aten::linalg_vector_norm`). Each successive workaround (hand-rolled L2 normalize, tensor-wrapped epsilon, skipping normalization entirely) just moves the error to the next op.

joe32140:
Fixes #93.

`pylate-onnx-export` only worked on ModernBERT-family checkpoints. Exporting `LiquidAI/LFM2.5-ColBERT-350M` or `jinaai/jina-colbert-v2` failed at several independent points — and two further bugs produced a model that exported *"successfully"* but was silently wrong or unusable at inference.

Bugs 1–5 correspond to the ones reported in #93; 6–8 surfaced only after the earlier ones were cleared.

## Bugs fixed

| # | Symptom | Cause / fix |
|---|---------|-------------|
| 1 | `KeyError: 'token_type_ids'` on any non-ModernBERT arch | `detect_model_architecture()` hardcoded a `"ModernBert"` class-name allowlist; everything else was assumed to use `token_type_ids`. Now probes the tokenizer directly. |
| 2 | `torch.export` fails on LFM2 | LFM2 defaults `use_cache=True`, building a KV cache and hitting data-dependent control flow (`cache_position[0] > 0`) that `torch.export` can't trace. Force `use_cache=False` where supported. |
| 3 | Exported model won't load in ORT | LFM2/Jina load in **bfloat16**; ORT's CPU EP rejects bf16 for several ops (e.g. `Conv`). The model passed `check_model` but failed at session creation. Cast to fp32 before export. |
| 4 | Jina-v2 can't be loaded at all | `trust_remote_code` was never forwarded. Added as a param + `--trust-remote-code` flag. |
| 5 | `No module named 'onnxscript'` / `einops` | Undeclared deps (`onnxscript` for torch's dynamo exporter, `einops` for Jina's custom modeling code). |
| 6 | **Batch dim silently baked static** | `dynamic_axes` is the legacy `dynamo=False` API and is not honored under the current `dynamo=True` default. LFM2's causal mask specialized to `batch_size=1`; inference then failed on *every* other batch size. Switched to `dynamic_shapes` + a 2-row dummy input. |
| 7 | **Query expansion silently disabled** | `do_query_expansion=False` was hardcoded, overriding the trained config. It's baked into `onnx_config.json` and acted on by the Rust side, so MASK-token query expansion was disabled for models trained with it (LFM2.5, Jina-v2). Invisible until now because `GTE-ModernColBERT-v1` also defaults to `False`. Now left to pylate to resolve. |
| 8 | Verification crash on large models | `onnx.load()` + `check_model()` on the in-memory proto exceeds protobuf's 2 GB serialization limit for Jina-v2's ~2.2 GB fp32 weights — failing on a model that is valid on disk. Check by path instead. |

Bugs 6 and 7 are the notable ones: both produce an export that *looks* successful. 6 fails loudly but only at batch > 1; 7 never fails at all, it just quietly degrades retrieval.

## Verification

Exported models were checked against pylate's own forward pass — not just "the export ran":

| Check | Result |
|---|---|
| Document encoding (Rust ONNX vs pylate native) | max abs diff **1.8e-6** |
| Query encoding (Rust ONNX vs pylate native) | max abs diff **9.2e-7** |
| Projection head vs checkpoint `linear.weight` | **bitwise identical** |

`GTE-ModernColBERT-v1` still exports unchanged (no regression), and the existing suite passes (`15 passed`), `ruff check` + `format --check` clean.

## End-to-end sanity check (LFM2.5, SciFact)

Full BEIR SciFact test queries (300) against a 1000-doc subset (all 283 qrels-relevant docs + 717 random distractors), encoded through the Rust `next-plaid-onnx` crate, brute-force MaxSim:

| Metric | LFM2.5 |
|---|---|
| nDCG@10 | 0.764 |
| nDCG@100 | 0.787 |
| Recall@10 | 0.856 |
| Recall@100 | 0.967 |
| MAP | 0.736 |

This is a **pipeline sanity check, not a quality claim** — it's a reduced-distractor subset, so it is not comparable to published full-corpus leaderboard numbers (BM25 scores 0.777 on this same subset; SciFact is a notoriously BM25-friendly dataset). The numerical-parity figures above are the actual correctness evidence.

## Note on jina-colbert-v2

Jina-v2 now exports and runs correctly, but scores unexpectedly low on SciFact (0.439 nDCG@10 on the full 5183-doc corpus — below BM25's 0.652 measured through the same harness). I could not attribute this to the export: ONNX output is numerically identical to pylate's own forward pass, so pylate reproduces the same behavior. The eval harness itself was validated against BM25 ground truth (0.652 vs published ~0.665), and a truncation hypothesis was tested and refuted (raising `document_length` 300→512 moved nDCG@10 by −0.003).

That points to a pylate↔Jina compatibility issue upstream of this PR rather than an export defect, and I did not chase it further. Flagging it in case it's of interest — happy to open a separate issue.

### DIFF
--- a/next-plaid-onnx/python/pyproject.toml
+++ b/next-plaid-onnx/python/pyproject.toml
@@ -34,9 +34,11 @@ dependencies = [
     "transformers>=4.30.0",
     "safetensors>=0.4.0",
     "onnx>=1.14.0",
+    "onnxscript>=0.1.0",
     "onnxruntime>=1.16.0",
     "numpy>=1.24.0",
     "huggingface-hub>=0.20.0",
+    "einops>=0.7.0",
 ]
 
 [project.optional-dependencies]
--- a/next-plaid-onnx/python/src/colbert_export/cli.py
+++ b/next-plaid-onnx/python/src/colbert_export/cli.py
@@ -80,6 +80,12 @@ Supported models:
     )
 
     parser.add_argument(
+        "--trust-remote-code",
+        action="store_true",
+        help="Allow loading HF models that ship custom modeling code (e.g. jinaai/jina-colbert-v2)",
+    )
+
+    parser.add_argument(
         "--version",
         action="version",
         version="%(prog)s 1.6.5",
@@ -99,6 +105,7 @@ Supported models:
             quantize=not args.no_quantize,
             verbose=not args.quiet,
             force=args.force,
+            trust_remote_code=args.trust_remote_code,
         )
 
         # Push to Hub if requested
--- a/next-plaid-onnx/python/src/colbert_export/export.py
+++ b/next-plaid-onnx/python/src/colbert_export/export.py
@@ -12,6 +12,7 @@ IMPORTANT: Uses pylate's tokenizer which adds [Q] and [D] as special tokens.
 The ONNX model will have extended embeddings to support these tokens.
 """
 
+import inspect
 import json
 from pathlib import Path
 from typing import Optional
@@ -31,10 +32,12 @@ def detect_model_architecture(pylate_model: pylate_models.ColBERT) -> dict:
     auto_model = pylate_model[0].auto_model
     model_class_name = auto_model.__class__.__name__
 
-    # Check if model uses token_type_ids
-    uses_token_type_ids = True
-    if "ModernBert" in model_class_name:
-        uses_token_type_ids = False
+    # Whether the tokenizer actually emits token_type_ids varies by model
+    # family (e.g. Llama/LFM2-style tokenizers never produce it), so ask the
+    # tokenizer directly rather than guessing from the architecture name.
+    tokenizer = pylate_model[0].tokenizer
+    probe = tokenizer("probe", return_tensors="pt")
+    uses_token_type_ids = "token_type_ids" in probe
 
     # Get hidden size and output dimension
     config = auto_model.config
@@ -73,6 +76,12 @@ class ColBERTForONNX(nn.Module):
 
         self.uses_token_type_ids = uses_token_type_ids
 
+        # Decoder-style backbones (e.g. LFM2) default use_cache=True from their
+        # config, which builds a KV cache and triggers data-dependent control
+        # flow (cache_position[0] > 0) that torch.export can't trace. We only
+        # ever run a single forward pass here, so force it off when supported.
+        self.supports_use_cache = "use_cache" in inspect.signature(self.bert.forward).parameters
+
     def forward(
         self,
         input_ids: torch.Tensor,
@@ -84,16 +93,19 @@ class ColBERTForONNX(nn.Module):
         Returns per-token embeddings [batch_size, seq_len, embedding_dim].
         """
         # Get hidden states from transformer
+        extra_kwargs = {"use_cache": False} if self.supports_use_cache else {}
         if self.uses_token_type_ids and token_type_ids is not None:
             outputs = self.bert(
                 input_ids=input_ids,
                 attention_mask=attention_mask,
                 token_type_ids=token_type_ids,
+                **extra_kwargs,
             )
         else:
             outputs = self.bert(
                 input_ids=input_ids,
                 attention_mask=attention_mask,
+                **extra_kwargs,
             )
         hidden_states = outputs.last_hidden_state
 
@@ -114,6 +126,7 @@ def export_model(
     quantize: bool = False,
     verbose: bool = True,
     force: bool = False,
+    trust_remote_code: bool = False,
 ) -> Path:
     """Export a ColBERT model from HuggingFace to ONNX format.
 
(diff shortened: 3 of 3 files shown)


raphaelsty:
Merged, thank you @joe32140

PR #180, opened 2026-08-24
feat(colgrep): measure whether colgrep actually saves tokens, with `colgrep ab`

raphaelsty:
## What

An opt-in experiment that answers the question users actually have — **does an agent with colgrep spend fewer tokens finding code than one without?** — plus a `colgrep ab` command that states the answer without requiring any statistics vocabulary.

Off by default. Search behaviour and output are untouched: `commands/search.rs` is byte-identical to `main`.

## How it works

With `colgrep settings --ab-test` on, each Claude Code session is randomly assigned an arm:

- **with colgrep** — the SessionStart hook injects the usual instructions
- **without** — the hook stays silent, so the agent uses native Grep/Glob exactly as if colgrep were never installed

Neither arm is told it is being measured; telling the model would change its behaviour and break the comparison. A new SessionEnd hook reads the transcript afterwards and adds up what locating code actually cost. The static `Grep|Glob` echo reminder becomes `colgrep --grep-hook` so control sessions never see colgrep mentioned anywhere.

## What is measured, and why

Every byte a search tool returned — colgrep, `grep`, `rg`, Grep/Glob, `find`, `cat` — **plus every byte a follow-up `Read` returned.**

Reads count on purpose. colgrep prints compact `path:lines`, which is cheap by itself but pushes the agent into opening files that grep's inline match content would have shown directly. Counting searches alone hides that cost, and it is real: in one measured session colgrep did a single 2,954-token search and then spent 10,876 tokens reading files.

The headline is **tokens per file:line reference found**, not tokens spent, because cost alone rewards a tool that returns nothing. If colgrep needs three calls to replace one `grep`, that shows up here — which is the entire point.

Whole-session token totals are deliberately *not* the metric: measured across 24 real sessions they were **99.3% cache re-reads of conversation history**, which drowns the search signal completely (the interval on that metric was a useless [0.72, 1.72]).

## The report

```
Is colgrep saving you tokens?

/Users/me/projects/myapp
18 sessions: 9 with colgrep, 9 without

  WORTH IT   finding code costs 59% less with colgrep
             the real saving is between 9% and 91% less

  per session                 with colgrep      without
  tokens used to find code           4,981        5,070   same
  file:line references found          15.5          6.5   2.4x more
  ──────────────────────────────────────────────────────────────────
  tokens per reference found           321          780   59% less

  searches run                         4.0          3.5
  files opened afterwards              1.2          1.0
  share of codebase read             0.30%        0.29%

  Both used about the same number of tokens looking, but colgrep found 2.4x
  more references, so each one came cheaper.

What the rows mean
  tokens used to find code    everything a search returned (colgrep, grep, rg, Grep/Glob),
                              plus every file the agent opened afterwards to check a result
  file:line references found  distinct file:line spots the agent cited in its answers —
                              a stand-in for how much of the answer it actually located
  tokens per reference found  the first row divided by the second — the number the verdict
                              is based on. A tool that needs three searches to replace one
                              grep costs more per reference even if each search is cheap.
```

Design points worth knowing, since they were all arrived at by finding the earlier version wrong:

- **The verdict comes first** (`WORTH IT` / `NOT HERE` / `TOO EARLY`) with its interval, so nobody has to hunt for the conclusion.
- **The table shows the arithmetic.** The first two rows divide into the third and you can check it by hand: 4,981 ÷ 15.5 = 321. An earlier draft put a per-session median next to a cross-session total, so the division it invited was simply wrong (5,041 ÷ 62 = 81, not 321).
- **Per-session averages, not totals**, so that division stays exact even when one arm ran more sessions than the other — there is a test pinning that under deliberately unbalanced arms.
- **A generated sentence names the mechanism**, adapted to the data, so a result like "same tokens, 59% cheaper" cannot read as a contradiction.
- **A legend prints once** at the end, so no row requires prior knowledge.
- **The p-value is demoted** to a dimmed footnote. A dashboard you can re-run at will invites reading a bare "significant" the moment it appears, which is not what a fixed-sample p-value means; the interval carries the honesty instead. Bootstrap intervals are seeded from the data, so numbers never jitter between runs.

`colgrep ab [PATH] [--json]`.

## Replayed against real sessions

Replaying 24 real agent sessions (12 paired tasks across `trl`, `accelerate`, `datasets`) through the shipped pipeline gives an honest mix rather than a uniform win:

| project | tokens per reference (with / without) | verdict |
|---|---|---|
| trl | 321 / 780 | ✓ 59% less (9%–91%) |
| accelerate | 275 / 376 | ? too early (63% less – 34% more) |
| datasets | 726 / 533 | ✗ 36% more |

That spread is the feature working. A tool that only ever reported wins would not be worth having.

## What this replaces, and why

An earlier revision of this branch shipped a second experiment that priced each colgrep search against a *simulated* `grep` of the same query, and reported 48–69% savings per search. **That has been removed**, because checking it against real transcripts showed the baseline was not a fair opponent:

- **82% of the greps an agent actually writes are piped to `head`**, and every one targets a specific file or directory
- the simulation grepped the whole indexed corpus, unbounded

So it compared colgrep against a strawman nobody uses. Worse, the unit was wrong: per-search savings cannot see how many calls each tool needs, and that is what decides the answer. Deleting `abtest/baseline.rs` also removes all sampling machinery from the search path, which is why `search.rs` is unchanged now.

## Review notes

- **Behaviour change:** the `Grep|Glob` PreToolUse reminder now spawns the colgrep binary instead of `echo`, and `hook.json` gains a `SessionEnd` entry. Users must rerun `colgrep --install-claude-code` once.
- **Statistics:** rank-based only (heavy-tailed data), hand-rolled with tie and continuity corrections plus a deterministic xorshift bootstrap — no new dependency. `abtest/stats_math.rs`, with tests against known quantiles, separated/interleaved samples, a determinism check, and the English-phrasing helpers (generated prose is easy to break silently).
- **Schema versioning:** `SAMPLE_VERSION` is 2 and older rows are dropped on load, since the measured quantities changed and mixing them would corrupt the report.
- **Contamination is detected:** a control session that ran colgrep anyway (someone asked for it) is counted and flagged rather than silently averaged in.
- **`colgrep ab` is read-only** — verified by checksumming the whole data directory before and after; it also never triggers the index auto-update a search would, and takes ~10ms.
- **Known gaps, deliberately not addressed:** repeated peeking at a re-runnable dashboard inflates false positives (the rigorous fix is a confidence sequence), and per-project verdicts are not corrected for multiplicity. Showing intervals instead of binary claims blunts both; neither is solved.
- 787 tests green, clippy clean, `cargo fmt` clean.
- First commit (`a2675b8`) is an unrelated pre-existing CI fix in `next-plaid-api`; see the earlier comment on this PR.


raphaelsty:
### CI fix — heads-up on the second commit

CI was red on two jobs; both are now green.

**1. `Documentation` — my bug.** `cargo doc` runs with `RUSTDOCFLAGS: -D warnings`, and a doc comment of mine read ``"token:<word>"``, which rustdoc parses as an unclosed HTML tag (`rustdoc::invalid-html-tags`). Fixed by making it a code span. Amended into the feature commit since it's a typo in this PR's own new code.

**2. `Clippy` — pre-existing, unrelated to this PR** (hence the separate commit `a1d7a4a` touching `next-plaid-api/src/models.rs`, the one non-colgrep file in the diff).

The workspace clippy job failed on `decode_b64_embeddings` using `chunks_exact(4)`, which trips `clippy::chunks_exact_to_as_chunks` — a lint that did not exist when that code was written (it landed in Rust 1.98; that line is from #155, July). This PR changes nothing in `next-plaid-api`, and the same failure is latent on `main` — it went unnoticed because `main`'s recent CI runs only exercised the Homebrew tap job. The fix is clippy's own suggested `as_chunks::<4>()`, which also removes the manual `[c[0], c[1], c[2], c[3]]` indexing since chunks arrive as `[u8; 4]`; the pre-existing length check guarantees an empty remainder. All 56 `next-plaid-api` tests pass.

Worth noting for whoever merges: that lint will hit any other PR opened against `main` until this lands, so it may be worth cherry-picking `a1d7a4a` to `main` independently if this PR sits in review.

One transparency note on my verification: I could not reproduce the clippy failure locally — this machine's stable toolchain is 1.91 and nightly is 1.93, so neither has the 1.98 lint. I did not bump the toolchain to check; the fix is the compiler's verbatim suggestion and CI has now confirmed it.

PR #183, opened 2026-09-22
fix(fts): stop re-running the MATCH once per filtered id

raphaelsty:
## The bug

`text_search::search_filtered` narrows a full-text search with

```sql
... WHERE METADATA_FTS MATCH ? AND rowid IN (<subset>)
```

That shape reads like a pre-filter and behaves like the opposite. SQLite hands the rowid list to FTS5's `xBestIndex` as an **equality** constraint and then drives the loop from the ids, re-running the whole full-text query once per id at ~100 µs a go. The plans differ by one character:

```
no filter :  SCAN METADATA_FTS VIRTUAL TABLE INDEX 0:M1     →     1 ms
filtered  :  SCAN METADATA_FTS VIRTUAL TABLE INDEX 0:=M1    →  5908 ms
             USING ROWID SEARCH ON TABLE t1 FOR IN-OPERATOR
```

So filtering cost O(subset), and got *slower* the more documents the filter let through — the opposite of what a pre-filter is for. Chunking the ids to stay under `SQLITE_PARAM_LIMIT` only spread the same per-id work across more statements.

## The fix

Run the MATCH exactly once and drop non-subset rows as they stream in. A filtered search is then bounded by the cost of the unfiltered search it refines, whatever the subset size.

## Numbers

Measured on a 685 k-document index (21 M embeddings, 979 MB `metadata.db`), query `"retrieval" "augmented" "generation"`, `top_k=600`:

| subset | before | after |
|---|---|---|
| none (baseline) | 16 ms | 16 ms |
| 41 k (one source) | 3 674 ms | 3 ms |
| 68 k (two sources) | 6 523 ms | 5 ms |
| 499 k | 37 291 ms | 22 ms |

Across 20 (query × subset) combinations on that index the new path returns the **identical id sequence**, 43×–1595× faster. Worst case before was `twitter` (499 k ids) × `"model"` at 205 s.

## Ranking

Unchanged: the result is the unfiltered ranking restricted to the subset. Document id is added as a tiebreak so equal-scoring rows come back in a stable order across calls.

## Tests

- `test_search_filtered_large_subset_matches_unfiltered_ranking` — a subset spanning several old chunks returns exactly `search()` filtered down to it, in order, with scores descending.
- `test_search_filtered_subset_edges` — an all-admitting filter matches `search()`, unknown ids return nothing, `top_k = 0` returns nothing.

Full suite green (`cargo test -p next-plaid`), `cargo clippy --all-targets` and `cargo fmt --check` clean.

## Note for later

Two neighbouring costs this PR does not touch, both O(subset) per query:

- `filtering::where_condition` is a full `SCAN METADATA` — the table carries no secondary indexes, so resolving `source IN ('github')` on that index takes ~155 ms however selective it is.
- `search::search_one_mmap` walks every subset document's codes to build `eligible_centroids`, and scales `n_ivf_probe` by `num_docs / subset_len`. Together they take the PLAID half from 113 ms unfiltered to 516 ms behind a 499 k-document filter — real, but a rounding error next to the FTS problem above.


raphaelsty:
Pushed a second commit, `fix(api): as_chunks instead of chunks_exact in decode_b64_embeddings`, which is unrelated to the FTS work. `clippy::chunks_exact_to_as_chunks` on current stable fails `next-plaid-api` under `-D warnings`, so the Clippy job is red for every PR until it goes — including this one. Same one-line change already carried downstream in knowledge-api.

